PR #6, opened 2017-10-03
Replace enums with defines in API

jciupis:


src/nrf_drv_radio802154.h
```diff
@@ -53,39 +53,36 @@ extern "C" {
...
-typedef enum
-{
-    NRF_DRV_RADIO802154_TX_ERROR_BUSY_CHANNEL,    /**< CCA reported busy channel prior to transmission. */
-    NRF_DRV_RADIO802154_TX_ERROR_INVALID_ACK,     /**< Received ACK frame is other than expected. */
-    NRF_DRV_RADIO802154_TX_ERROR_NO_MEM,          /**< No receive buffer are available to receive an ACK. */
-    NRF_DRV_RADIO802154_TX_ERROR_TIMESLOT_ENDED,  /**< Radio timeslot ended during transmission procedure. */
-} nrf_drv_radio802154_tx_error_t;
+typedef uint8_t nrf_drv_radio802154_tx_error_t;
+
+#define NRF_DRV_RADIO802154_TX_ERROR_BUSY_CHANNEL      0x11 /**< CCA reported busy channel prior to transmission. */
```
hubertmis: Is there a reason this define sections starts with 0x11 instead of 0x00?
It also applies to other define sections.

LuDuda: Agree, i would start iteration from 0x00.

jciupis: Error codes start with 1 instead of 0 because I thought that 0x00 is usually return code for success and this way it is clear that all these return codes are errors. The second digit is different for every group of defines so that they are easily recognizable i.e. when debugging.

hubertmis: I agree we can start error codes with 1. But I would not keep "error space" in error code. You can easily recognize error code by function that is called.

jciupis: Agree, fixed.

PR #13, opened 2017-11-22
Provide a way to pass custom configuration to the driver.

hubertmis:


hubertmis:
I've found a few more labels that needed to be updated. Now should be OK.

PR #16, opened 2017-12-04
Remove dependency to softdevice.h file.

LuDuda:
Currently softdevice.h is provided from OpenThread and our test framework.

This PR removes unnecessary dependecy to this file and includes only one needed SoftDevice header.


LuDuda:
Yes i will report. Thanks.

PR #18, opened 2017-12-08
Use PHYEND event instead of END event

rlubos:
This PR introduces usage of PHYEND event in 802.15.4 transmission/reception.

Because this event is not present in older revision of the nRF52840 chip, some additional logic had to be added to discriminate chip revision in runtime.


src/nrf_drv_radio802154_fsm.c
```diff
@@ -50,28 +50,41 @@
...
 #include "nrf_drv_radio802154_priority_drop.h"
 #include "nrf_drv_radio802154_procedures_duration.h"
+#include "nrf_drv_radio802154_revision.h"
 #include "nrf_drv_radio802154_rx_buffer.h"
 #include "fem/nrf_fem_control_api.h"
 #include "hal/nrf_radio.h"
 #include "raal/nrf_raal_api.h"
 
+/// Workaround for missing PHYEND event in older chip revision.
+static inline uint32_t short_phyend_disable_mask_workaround(void)
```
hubertmis: I would prefer to make this compile-time instead of run-time. It's highly unlikely to build a firmware image that can be run simultanously on more than 1 nRF52 device types.

src/nrf_drv_radio802154_fsm.c
```diff
@@ -50,28 +50,41 @@
...
+    }
+
+    return NRF_RADIO_SHORT_END_DISABLE_MASK;
+}
+
 /// Value set to SHORTS register when no shorts should be enabled.
 #define SHORTS_IDLE         0
 /// Value set to SHORTS register when receiver is waiting for incoming frame.
-#define SHORTS_RX_INITIAL   (NRF_RADIO_SHORT_END_DISABLE_MASK | NRF_RADIO_SHORT_DISABLED_TXEN_MASK | \
+#define SHORTS_RX_INITIAL   (short_phyend_disable_mask_workaround() | NRF_RADIO_SHORT_DISABLED_TXEN_MASK | \
```
hubertmis: In RX state we should use END->DISABLE instead of PHYEND->DISABLE

src/nrf_drv_radio802154_fsm.c
```diff
@@ -50,28 +50,41 @@
...
                              NRF_RADIO_SHORT_FRAMESTART_BCSTART_MASK)
+
 /// Value set to SHORTS register when receiver started receiving a frame.
-#define SHORTS_RX_FOLLOWING (NRF_RADIO_SHORT_END_DISABLE_MASK | NRF_RADIO_SHORT_READY_START_MASK | \
+#define SHORTS_RX_FOLLOWING (short_phyend_disable_mask_workaround() | NRF_RADIO_SHORT_READY_START_MASK | \
```
hubertmis: Note: here PHYEND->DISABLE short is used to disable transmitter after Ack transmission. PHYEND is correct here.

src/nrf_drv_radio802154_fsm.c
```diff
@@ -904,7 +924,7 @@ static inline void irq_bcmatch_state_rx_header(void)
 /** This event is generated when radio receives invalid frame with length set to 0 or started
  *  receiving a frame after other procedure locked mutex.
  */
-static inline void irq_end_state_waiting_rx_frame(void)
+static inline void irq_phyend_state_waiting_rx_frame(void)
```
hubertmis: According to spec it should be END event (when leaving RX state)

src/nrf_drv_radio802154_fsm.c
```diff
@@ -924,7 +944,7 @@ static inline void irq_end_state_waiting_rx_frame(void)
 }
 
 /// This event is generated if frame being received is shorter than expected MAC header length.
-static inline void irq_end_state_rx_header(void)
+static inline void irq_phyend_state_rx_header(void)
```
hubertmis: END instead of PHYEND

src/nrf_drv_radio802154_fsm.c
```diff
@@ -933,7 +953,7 @@ static inline void irq_end_state_rx_header(void)
 }
 
 /// This event is generated when radio peripheral ends receiving of a complete frame.
-static inline void irq_end_state_rx_frame(void)
+static inline void irq_phyend_state_rx_frame(void)
```
hubertmis: END again

src/nrf_drv_radio802154_fsm.c
```diff
@@ -982,15 +1002,15 @@ static inline void irq_end_state_rx_frame(void)
 }
 
 /// This event is generated when the radio ends transmission of ACK frame.
-static inline void irq_end_state_tx_ack(void)
+static inline void irq_phyend_state_tx_ack(void)
 {
     assert(nrf_radio_shorts_get() == SHORTS_TX_ACK ||
-           nrf_radio_shorts_get() == SHORTS_RX_FOLLOWING); // In case END is handled before READY
+           nrf_radio_shorts_get() == SHORTS_RX_FOLLOWING); // In case PHYEND is handled before READY
```
hubertmis: It seems this is not needed anymore. We can move handling PHYEND after READY handler (and still keep END handler in the place it is right now).

src/nrf_drv_radio802154_fsm.c
```diff
@@ -1000,18 +1020,18 @@ static inline void irq_end_state_tx_ack(void)
 /** This event may occur at the beginning of transmission procedure (the procedure already has
  *  disabled shorts).
  */
-static inline void irq_end_state_cca_before_tx(void)
+static inline void irq_phyend_state_cca_before_tx(void)
```
hubertmis: END (it's end of RX event)

src/nrf_drv_radio802154_fsm.c
```diff
@@ -1031,7 +1051,7 @@ static inline void irq_end_state_tx_frame(void)
 }
 
 /// This event is generated when the radio ends receiving of ACK frame.
-static inline void irq_end_state_rx_ack(void)
+static inline void irq_phyend_state_rx_ack(void)
```
hubertmis: It's receiving of an ACK frame. It should generate only END event, no PHYEND

src/nrf_drv_radio802154_fsm.c
```diff
@@ -1448,40 +1468,49 @@ static inline void irq_handler(void)
         nrf_drv_radio802154_log(EVENT_TRACE_EXIT, FUNCTION_EVENT_BCMATCH);
     }
 
-    if (nrf_radio_event_get(NRF_RADIO_EVENT_END))
+    if ((nrf_drv_radio802154_revision_has_phyend_event() && nrf_radio_event_get(NRF_RADIO_EVENT_PHYEND)) ||
```
hubertmis: I would handle END and PHYEND separately

src/nrf_drv_radio802154_revision.h
```diff
@@ -0,0 +1,83 @@
+/* Copyright (c) 2017, Nordic Semiconductor ASA
```
hubertmis: Instead of this file I suggest compile flag. I think there is one in SDK like NRF52840_AAAA or sth like this.

src/nrf_drv_radio802154_fsm.c
hubertmis: Second thought: it's a bad idea to move PHYEND after READY. It should be placed right after END.

src/nrf_drv_radio802154_fsm.c
LuDuda: Don't you think there will be too many bad consequences?

E.g. two OpenThread libraries in the SDK, additional documentation in OpenThread, Zephyr and other our internal projects. There is requirement to support both AA and BA+ chips because of the huge number Preview Kits already distributed. What if developer has mix of PCA10056 0.9 and 0.10 on the desk? Generally i think its bad idea to make it compile-time in the context of other projects (not in the context of stand alone driver project).

LuDuda: Alternatively, maybe we can add additional define, by default set to 0 - and then run-time chip version check will be made, or set to 1 and then there will be requirement to specify the version of the chip compile-time.

hubertmis: That's right that development process will be harder with compile time selection. But time and power efficiency will be improved in products. I believe we should provide as efficient driver as we can to show ultra low power consumption of our chip. 

Let's compromise. We can provide compile time configuration that specifies board version or run time decision. For development build user can use run time decision but for production release compile time option should be used.

This will make code harder to maintain and 2x more test configurations for us but it will provide efficient driver for products and easy to use for development.

hubertmis: Oh, I can see you already suggested my compromise :-)

hubertmis: Or we can make version AAAA default. According to our experience it may be run on both chip revisions. Only some features do not work as expected.

src/nrf_drv_radio802154_revision.h
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * @brief Internal auxiliary function to check if the program is running on NRF52840 version
+ *        that supports PHYEND event.
+ *
+ * @retval true  If PHYEND event is supported.
+ * @retval false If PHYEND event is not supported.
+ */
+static inline bool nrf_drv_radio802154_revision_has_phyend_event(void)
+{
+    return !nrf_drv_radio802154_revision_type_52840_proto1();
```
hubertmis: In case we decide to use run-time configuration I would optimize number of instructions instead of memory usage:
Provide static enum containing current chip revision. Fill this enum on module initialization. Then this function can check single enum value instead of a few volatile variables + logical operations.

LuDuda: +1

src/nrf_drv_radio802154_fsm.c
rlubos: Double configuration sounds like a good compromise for me. But I think we should stick to run ime selection as a default, as I would prioritise maximum out-of-the-box feature set over power efficency. We should document though, that specific chip revision should be selected compile time, for maximum performance/power efficiency.

src/nrf_drv_radio802154_fsm.c
rlubos: Perhaps it should, but it appears that both events are raised in both RX and TX, so I wanted to minimise changes in the driver itself. But it can be reworked though.

src/nrf_drv_radio802154_fsm.c
```diff
@@ -468,13 +480,21 @@ static void nrf_radio_init(void)
     nrf_radio_mhmu_pattern_mask_set(MHMU_MASK);
 
     nrf_radio_int_enable(NRF_RADIO_INT_FRAMESTART_MASK);
-    nrf_radio_int_enable(NRF_RADIO_INT_END_MASK);
```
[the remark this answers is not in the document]
rlubos: Spec is very vauge in terms of this new event. Actual test showed that it is raised in both, TX and RX, while in RX both events strike almost simultaneously.

src/nrf_drv_radio802154_fsm.c
hubertmis: If that is true, we should confirm it with IC team how does it work.
It is possible that PHYEND event in RX is raised when transmission is received, while END is raised a little later when DMA ends it's operation.

src/nrf_drv_radio802154_revision.h
rlubos: Sounds good.

src/nrf_drv_radio802154_fsm.c
rlubos: Well the time difference between END and PHYEND was near to my logic analyser resoultion, so I assumed that they are raised at the same time. But it was a surprise for me too that both events are raised in RX. I agree though that it would be better to stick to the END event in the next iteration.

hubertmis: I suggest to postpone my review until we get more info from IC team. If we could use PHYEND instead of END in RX it would not complicate FSM even more.

src/nrf_drv_radio802154.c
```diff
@@ -115,6 +116,7 @@ int8_t nrf_drv_radio802154_dbm_from_energy_level_calculate(uint8_t energy_level)
 
 void nrf_drv_radio802154_init(void)
 {
+    nrf_drv_radio802154_revision_init();
```
hubertmis: I would keep it in alphabetic order. Or is there any case to keep it in this order?

src/nrf_drv_radio802154_fsm.c
```diff
@@ -50,28 +50,50 @@
...
+    if (nrf_drv_radio802154_revision_has_phyend_event())
+    {
+        return NRF_RADIO_SHORT_PHYEND_DISABLE_MASK;
+    }
+
+    return NRF_RADIO_SHORT_END_DISABLE_MASK;
+}
+
+#define SHORT_PHYEND_DISABLE_MASK short_phyend_disable_mask_get()
+#endif
```
hubertmis: Please add `// NRF52840_AAAA` comment in this line

src/nrf_drv_radio802154_fsm.c
```diff
@@ -50,28 +50,50 @@
...
 #include "nrf_drv_radio802154_priority_drop.h"
 #include "nrf_drv_radio802154_procedures_duration.h"
+#include "nrf_drv_radio802154_revision.h"
 #include "nrf_drv_radio802154_rx_buffer.h"
 #include "fem/nrf_fem_control_api.h"
 #include "hal/nrf_radio.h"
 #include "raal/nrf_raal_api.h"
 
+
+/// Workaround for missing PHYEND event in older chip revision.
```
hubertmis: I like this workaround very much :+1:

src/nrf_drv_radio802154_revision.c
```diff
@@ -0,0 +1,108 @@
...
+    }
+    else
+    {
+        m_nrf52840_revision = NRF52840_REVISION_UNKNOWN;
+    }
+}
+
+bool nrf_drv_radio802154_revision_has_phyend_event(void)
+{
+    return m_nrf52840_revision != NRF52840_REVISION_AAAA;
```
hubertmis: I would prefer switch-case here. I think it would be cleaner.

src/nrf_drv_radio802154.c
rlubos: I've put it intentionally at the beginning, in case some other modules had some revision-specific behavior. It might not be a case for now, but perhaps it might be an issue in the feature. What do you think?

src/nrf_drv_radio802154_fsm.c
rlubos: Is this still OK if there are multiple conditions? Why `NRF52840_AAAA` and not `NRF52840_AABA`?

src/nrf_drv_radio802154_revision.c
rlubos: Sounds good.

src/nrf_drv_radio802154.c
hubertmis: Probably we could put every single initialization at the beginning of the sequence with the same reason. I prefer alphabetical order until we have a specific case that we need to reorder it. When this specific case happens, we would add a comment why we used specific order.

src/nrf_drv_radio802154_fsm.c
hubertmis: It was the first one :)
Maybe something more descriptive then, like `// MCU_REVISION`?

src/hal/nrf_radio.h
```diff
@@ -89,6 +89,7 @@ typedef enum /*lint -save -e30 -esym(628,__INTADDR__) */
     NRF_RADIO_EVENT_FRAMESTART = offsetof(NRF_RADIO_Type, EVENTS_FRAMESTART), /**< IEEE 802.15.4 length field received. */
     NRF_RADIO_EVENT_EDEND      = offsetof(NRF_RADIO_Type, EVENTS_EDEND),      /**< Energy Detection procedure ended. */
     NRF_RADIO_EVENT_BCMATCH    = offsetof(NRF_RADIO_Type, EVENTS_BCMATCH),    /**< Bit counter reached bit count value. */
+    NRF_RADIO_EVENT_PHYEND     = offsetof(NRF_RADIO_Type, EVENTS_PHYEND),     /**< Generated in Ble_LR125Kbit, Ble_LR500Kbit and BleIeee802154_250Kbit modes when last bit is sent on air. */
```
LuDuda: nit: Ieee802154_250Kbit

src/nrf_drv_radio802154.c
rlubos: I'm not convinced, but I'll move as you request.

src/nrf_drv_radio802154_fsm.c
rlubos: :+1:

src/nrf_drv_radio802154.c
LuDuda: I lean towards to rlubos argumentation. But either way its okay for me.

src/nrf_drv_radio802154_revision.h
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+#ifndef NRF_DRV_RADIO802154_REVISION_H_
+#define NRF_DRV_RADIO802154_REVISION_H_
+
+#include <stdbool.h>
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
```
LuDuda: There is no

#ifdef __cplusplus
}
#endif

at the end.

src/hal/nrf_radio.h
hubertmis: This typo is on OPS. I'd leave it for consistency

LuDuda: I will inform our team about this issue. Personally I wouldn't duplicate a typo.

hubertmis: I'm OK to fix it before OPS is fixed

rlubos:
I've uploaded first version that can be considered final. It worked on my desk (as any other version uploaded here), but we should probably test it more thoroughly.

src/nrf_drv_radio802154_fsm.c
```diff
@@ -475,6 +496,18 @@ static void nrf_radio_init(void)
     nrf_radio_int_enable(NRF_RADIO_INT_READY_MASK);
     nrf_radio_int_enable(NRF_RADIO_INT_BCMATCH_MASK);
     nrf_radio_int_enable(NRF_RADIO_INT_EDEND_MASK);
+
+    /// Workaround for missing PHYEND event in older chip revision.
```
hubertmis: I heave an idea to make this workaround a little neater:
Instead of putting `#if ... #elif ... #else ... #endif` sequence in a few places in this file checking for PHYEND support, you can move preprocessor section to the `nrf_drv_radio802154_revision_has_phyend_event()` function. `#if` and `#elif` should contain simple `return true` or `return false`, while `#else` section should contain switch-case. With LTO output assembler should be the same with C code easier to maintain.

src/nrf_drv_radio802154_fsm.c
```diff
@@ -1448,50 +1481,104 @@ static inline void irq_handler(void)
...
-                irq_end_state_waiting_rx_frame();
-                break;
-
             case RADIO_STATE_RX_HEADER:
-                irq_end_state_rx_header();
-                break;
-
             case RADIO_STATE_RX_FRAME:
-                irq_end_state_rx_frame();
+            case RADIO_STATE_RX_ACK:
```
hubertmis: I would keep order of cases as it was. Each `switch (m_state)` in this file has (or should have) order of cases the same as order of labels in enum. I think this approach makes it easier to check how particular case is handled. I know that it makes a few more lines of code, but I think this waste is worth easier maintenance and readability.

src/nrf_drv_radio802154_fsm.c
```diff
@@ -1448,50 +1482,92 @@ static inline void irq_handler(void)
...
             case RADIO_STATE_TX_FRAME:
-                irq_end_state_tx_frame();
-                break;
-
-            case RADIO_STATE_RX_ACK: // Ended receiving of ACK.
-                irq_end_state_rx_ack();
+                irq_phyend_state_tx_frame();
                 break;
 
+            case RADIO_STATE_RX_ACK:
```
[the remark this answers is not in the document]
rlubos: That's how I've originally made it, @hubertmis requested to change it so enums are in order of their declaration.

PR #21, opened 2017-12-14
Use fast ramp up mode whenever possible

hubertmis:
Fast ramp cannot be used with hardware IFS support.
Performed smoke tests with Logic Analyzer. Looks correctly. Feature tests passed on my desk. I'll perform smoke, regression and feature tests on build server after merge.


hubertmis:
I've updated FEM configuration to make it work correctly with fast ramp up mode.

PR #26, opened 2018-01-03
Critical sections refactoring

hubertmis:
This PR allows to call driver requests from any IRQ priority.


KaSroka:
Regression test

KaSroka:
Regression test

KaSroka:
Regression test

KaSroka:
Regression test

src/nrf_drv_radio802154_critical_section.c
```diff
@@ -39,28 +39,112 @@
...
+}
+
+/** @brief Check if active vector priority is equal to RADIO IRQ priority.
+ *
+ * @retval true   Active vector priority is equal to RADIO IRQ priority.
+ * @retval false  Active vector priority is not equal to RADIO IRQ priority.
+ */
+static bool current_irq_priority_is_radio(void)
+{
+    return nrf_drv_radio802154_critical_section_active_vector_priority_get() ==
```
KaSroka: Wow :smile:

src/nrf_drv_radio802154_critical_section.c
```diff
@@ -71,53 +155,110 @@ void nrf_drv_radio802154_critical_section_enter(void)
...
+
     nrf_drv_radio802154_log(EVENT_TRACE_ENTER, FUNCTION_CRIT_SECT_EXIT);
 
 #if PREVENT_NESTED_CRIT_SECTIONS
     assert(m_in_critical_section);
     m_in_critical_section = false;
 #endif // PREVENT_NESTED_CRIT_SECTIONS
 
-    radio_state_t state;
+    if (exiting_mutex_lock())
```
KaSroka: What about bool exiting = exiting_mutex_lock(), In line 203 if (cnt == 1) -> if (cnt == 1 && exiting) and if (exiting) before 213. Then we could remove 215-224

src/nrf_drv_radio802154_critical_section.c
```diff
@@ -71,53 +155,110 @@ void nrf_drv_radio802154_critical_section_enter(void)
...
-    nrf_drv_radio802154_timer_critical_section_exit();
+            assert(cnt > 0);
+        }
+        while (__STREXB(cnt - 1, &m_nested_critical_section_counter));
+    }
 
     nrf_drv_radio802154_log(EVENT_TRACE_EXIT, FUNCTION_CRIT_SECT_EXIT);
 }
 
+void nrf_drv_radio802154_critical_section_nesting_allow(void)
```
KaSroka: I would merge nrf_drv_radio802154_critical_section_nesting_allow with nrf_drv_radio802154_critical_section_nesting_disallow:
```c
void nrf_drv_radio802154_critical_section_nesting_allow(bool allow)
{
    assert(m_nested_critical_section_is_allowed != allow);
    assert(m_nested_critical_section_counter > 0);

    m_nested_critical_section_is_allowed = allow;
}
 ```

File: src/nrf_drv_radio802154_critical_section.c [hunk 4]
```diff
@@ -71,53 +155,110 @@ void nrf_drv_radio802154_critical_section_enter(void)
...
+uint32_t nrf_drv_radio802154_critical_section_active_vector_priority_get(void)
+{
+    uint32_t  active_vector_id = (SCB->ICSR & SCB_ICSR_VECTACTIVE_Msk) >> SCB_ICSR_VECTACTIVE_Pos;
+    IRQn_Type irq_number;
+    uint32_t  active_priority;
+
+    // Check if this function is called from main thread.
+    if (active_vector_id == 0)
+    {
+        return false;
```
KaSroka: return false when return type is uint32_t

src/nrf_drv_radio802154_critical_section.c
```diff
@@ -71,53 +155,110 @@ void nrf_drv_radio802154_critical_section_enter(void)
...
+
+void nrf_drv_radio802154_critical_section_nesting_disallow(void)
+{
+    assert(m_nested_critical_section_is_allowed);
+    assert(m_nested_critical_section_counter > 0);
+
+    m_nested_critical_section_is_allowed = false;
+}
+
+uint32_t nrf_drv_radio802154_critical_section_active_vector_priority_get(void)
```
KaSroka: Is 'critical_section' necessary in this name?

src/nrf_drv_radio802154_critical_section.c
```diff
@@ -39,28 +39,112 @@
...
 #endif // RAAL_SOFTDEVICE
 
+static volatile uint8_t m_nested_critical_section_counter;
+static volatile bool    m_nested_critical_section_is_allowed;
+static volatile uint8_t m_critical_section_exit_mutex;
+
 #if PREVENT_NESTED_CRIT_SECTIONS
 /// Flag indicating if the driver entered critical section.
 static volatile bool m_in_critical_section;
+#else // PREVENT_NESTED_CRIT_SECTIONS
```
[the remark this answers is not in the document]
hubertmis: Removed

src/nrf_drv_radio802154_critical_section.c
```diff
@@ -39,28 +39,112 @@
...
+
+#define CMSIS_IRQ_NUM_VECTACTIVE_DIFF 16
+
 
 #if RAAL_SOFTDEVICE
 // When Softdevice is selected as arbiter critical sections should not be nested.
 #define PREVENT_NESTED_CRIT_SECTIONS 1
 #endif // RAAL_SOFTDEVICE
 
+static volatile uint8_t m_nested_critical_section_counter;
```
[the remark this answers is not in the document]
hubertmis: Perhaps compiler can optimize memory layout with 8-bit variables. Or am I wrong?

src/nrf_drv_radio802154_critical_section.c
```diff
@@ -39,28 +39,112 @@
...
+
+    return true;
+}
+
+/** @brief Check if active vector priority is equal to RADIO IRQ priority.
+ *
+ * @retval true   Active vector priority is equal to RADIO IRQ priority.
+ * @retval false  Active vector priority is not equal to RADIO IRQ priority.
+ */
+static bool current_irq_priority_is_radio(void)
```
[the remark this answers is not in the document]
hubertmis: :+1:

src/nrf_drv_radio802154_critical_section.c
```diff
@@ -71,53 +155,110 @@ void nrf_drv_radio802154_critical_section_enter(void)
...
-        __ISB();
-#endif // !RAAL_SOFTDEVICE
+        if (cnt > 0 && (!m_nested_critical_section_is_allowed || !current_irq_priority_is_radio()))
+        {
+            __CLREX();
+            result = false;
+            break;
+        }
+
+        nrf_raal_critical_section_enter();
```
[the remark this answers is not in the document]
hubertmis: I've modified implementations to make it safe. I'll add comment to API description that this may be called multiple times during single critical section.

src/nrf_drv_radio802154_critical_section.c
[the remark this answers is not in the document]
KaSroka: Global/static variables will be optimized but automatic variables will be 32 bit anyway (AFAIK)

src/nrf_drv_radio802154_critical_section.c
hubertmis: Whole API (internal and external) of this module is in pairs: `_enter()` `_exit()`, `_lock()` `_unlock()`. I'm not sure if `_allow()` should be exception. @LuDuda, thoughts?

src/nrf_drv_radio802154_critical_section.c
hubertmis: According to coding standard it is. We should prefix public function name with module name. I think we could move this function to external module, but I'm not sure how to call it and where to place it. For sure it's not critical-section-specific function

src/nrf_drv_radio802154_critical_section.c
hubertmis: Good catch!

src/nrf_drv_radio802154_critical_section.c
KaSroka: I just don't like disallow word in current implementation

hubertmis: deny, reject, forbid, prohibit, resist?

KaSroka:
Regression test

KaSroka:
Regression test

src/nrf_drv_radio802154_critical_section.c
LuDuda: +1 for deny

hubertmis: @KaSroka ?

KaSroka: +1 for deny

KaSroka:
Regression test

hubertmis:
I've just provided variable to store allowed priority of nested critical section. It is intended to allow notifications from other IRQ handlers than RADIO. It will be used by ACK timeout procedure.

src/nrf_drv_radio802154_critical_section.c
```diff
@@ -39,28 +39,130 @@
...
+{
+    return m_nested_critical_section_allowed_priority ==
+            active_priority_convert(
+                    nrf_drv_radio802154_critical_section_active_vector_priority_get());
+}
+
+void nrf_drv_radio802154_critical_section_init(void)
+{
+    m_nested_critical_section_counter          = 0;
+    m_nested_critical_section_allowed_priority = NESTED_CRITICAL_SECTION_DENY_PRIORITY;
```
KaSroka: This is very confusing, I would change define name to NESTED_CRITICAL_SECTION_ALLOWED_PRIORITY_NONE

src/nrf_drv_radio802154_critical_section.c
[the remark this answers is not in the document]
KaSroka: Ok

src/nrf_drv_radio802154_critical_section.c
KaSroka: ..._active_irq_priority_get?

src/nrf_drv_radio802154_critical_section.c
```diff
@@ -39,28 +39,130 @@
...
+
+/** @brief Convert active priority value to int8_t type.
+ *
+ * @param[in]  active_priority  Active priority in uint32_t format
+ *
+ * @return Active_priority value in int8_t format.
+ */
+static int8_t active_priority_convert(uint32_t active_priority)
+{
+    return active_priority == UINT32_MAX ? INT32_MAX : (int8_t)active_priority;
```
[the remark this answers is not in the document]
hubertmis: We have following possible values of allowed priority:
* -1 - not allowed
* range 0-7 - vector priority
* `_MAX` - main thread context
I use `int8_t` to be able to store all of above. `active_priority` argument may have following values:
* range 0-7
* `UINT32_MAX`
Above implementation should cover all cases. I can add an `assert()` to make sure.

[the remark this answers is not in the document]
KaSroka: So I would change INT32_MAX to INT8_MAX, also any input value higher than INT8_MAX would be wrapped (but as you mentioned this shouldn't happen due to 0-7 range). Is it future proof?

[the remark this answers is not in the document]
hubertmis: Right, I meant `INT8_MAX` here. It is not future proof. I cannot predict new designs architecture and implementation. Let's stick to known architectures and improve it when required.

hubertmis:
Regression test

hubertmis:
Regression test

PR #28, opened 2018-01-08
Add cmock variant to radio driver .packsc module

jciupis:
To make unit testing the driver module possible, changes in the project description file are necessary.


nRF-IEEE-802.15.4-radio-driver.packsc
```diff
@@ -21,31 +28,31 @@
...
-            "src/platform/clock/nrf_drv_radio802154_clock_sdk.c",
-            "src/platform/timer/nrf_drv_radio802154_timer_none.c"
-        ],
         "_variants": [
             {
                 "_attrs": [
                     "public"
                 ],
                 "_files": [
+                    "src/nrf_drv_radio802154.c",
```
hubertmis: It would be great to provide a variable containing these c files common for a few variants. I'm afraid syntax of this file does not allow to create such variables. Am I right?

LuDuda: It could be done in this way only if cmock variant will be moved to separate component. I'm afraid that for now we have to leave it as is.

jciupis:
Regression test

KaSroka:
Regression test

KaSroka:
Regression test

nRF-IEEE-802.15.4-radio-driver.packsc
```diff
@@ -6,10 +6,17 @@
             "src"
         ],
         "_headers": [
-            "src/nrf_drv_radio802154_config.h"
-        ],
-        "_links": [
-            "nrf52840_xxaa"
+            "src/nrf_drv_radio802154.h",
+            "src/nrf_drv_radio802154_ack_pending_bit.h",
+            "src/nrf_drv_radio802154_critical_section.h",
```
hubertmis: It seems that nrf_drv_radio802154_fsm_hooks is missing

PR #30, opened 2018-01-10
Move frame filtering feature to separated module with generic API

hubertmis:


src/nrf_drv_radio802154_pib.h
```diff
@@ -118,6 +125,13 @@ void nrf_drv_radio802154_pib_tx_power_set(int8_t dbm);
  */
 void nrf_drv_radio802154_pib_pan_id_set(const uint8_t * p_pan_id);
 
+/**
+ * @brief Get Extended Address of this device
+ *
+ * @returns Pointer to buffer containing Extended address (8 bytes, little-endian).
```
rlubos: Should `Extended` be capitalized here?

src/nrf_drv_radio802154_const.h
```diff
@@ -36,36 +36,51 @@
...
+#include <stdint.h>
+
 #define ACK_HEADER_WITH_PENDING      0x12  ///< First byte of ACK frame containing pending bit
 #define ACK_HEADER_WITHOUT_PENDING   0x02  ///< First byte of ACK frame without pending bit
 
 #define ACK_LENGTH                   5         ///< Length of ACK frame
 #define ACK_REQUEST_BIT              (1 << 5)  ///< Ack request bit
 #define ACK_REQUEST_OFFSET           1         ///< Byte containing Ack request bit (+1 for frame length byte)
 #define DEST_ADDR_TYPE_EXTENDED      0x0c      ///< Bits containing extended destination address type
+#define DEST_ADDR_TYPE_NONE          0x00      ///< Bits containing not present destination address type
```
rlubos: I see this is ordered in alpabetic order, but wouldn't it be more clear, to separate values from offset and mask?
What I mean is to list values first in alphabetic order, and than mask/offset. Thoughts?

src/mac_features/nrf_drv_radio802154_filter.c
```diff
@@ -0,0 +1,297 @@
...
+
+    case FRAME_TYPE_FRAGMENT:
+    case FRAME_TYPE_EXTENDED:
+        return true;
+    }
+
+    return false;
+}
+
+static bool dst_addressing_may_be_present(uint8_t frame_type)
```
rlubos: I would repharse this function name in a more question-like manner -> `may_dst_addressing_be_present`. Or do I miss something?

src/mac_features/nrf_drv_radio802154_filter.c
```diff
@@ -0,0 +1,297 @@
...
+#define PHR_CHECK_OFFSET           PHR_SIZE
+#define FCF_CHECK_OFFSET           (PHR_SIZE + FCF_SIZE)
+#define SHORT_ADDR_CHECK_OFFSET    (DEST_ADDR_OFFSET + SHORT_ADDRESS_SIZE)
+#define EXTENDED_ADDR_CHECK_OFFSET (DEST_ADDR_OFFSET + EXTENDED_ADDRESS_SIZE)
+
+static bool frame_type_and_version_filter(uint8_t frame_type, uint8_t frame_version)
+{
+    switch (frame_type)
+    {
+    case FRAME_TYPE_BEACON:
```
rlubos: FCS recommends to indent each `case` within `switch`.

src/mac_features/nrf_drv_radio802154_filter.c
```diff
@@ -0,0 +1,297 @@
...
+    }
+
+    return false;
+}
+
+static bool dst_addressing_may_be_present(uint8_t frame_type)
+{
+    switch (frame_type)
+    {
+    case FRAME_TYPE_BEACON:
```
rlubos: Indentation

src/mac_features/nrf_drv_radio802154_filter.c
```diff
@@ -0,0 +1,297 @@
...
+    return false;
+}
+
+static bool dst_addressing_end_offset_get_2006(const uint8_t * p_psdu,
+                                               uint8_t       * p_num_bytes,
+                                               uint8_t         frame_type)
+{
+    switch (p_psdu[DEST_ADDR_TYPE_OFFSET] & DEST_ADDR_TYPE_MASK)
+    {
+    case DEST_ADDR_TYPE_SHORT:
```
rlubos: Indentation

src/mac_features/nrf_drv_radio802154_filter.c
```diff
@@ -0,0 +1,297 @@
...
+    case DEST_ADDR_TYPE_EXTENDED:
+        *p_num_bytes = EXTENDED_ADDR_CHECK_OFFSET;
+        return true;
+
+    case DEST_ADDR_TYPE_NONE:
+        if (frame_type == FRAME_TYPE_BEACON)
+        {
+            switch (p_psdu[SRC_ADDR_TYPE_OFFSET] & SRC_ADDR_TYPE_MASK)
+            {
+            case SRC_ADDR_TYPE_SHORT:
```
rlubos: Indentation. No use to write it in every function, just remember to check through the file.

src/mac_features/nrf_drv_radio802154_filter.c
```diff
@@ -0,0 +1,297 @@
...
+
+            case SRC_ADDR_TYPE_EXTENDED:
+                *p_num_bytes = EXTENDED_ADDR_CHECK_OFFSET;
+                return true;
+
+            default:
+                return false;
+            }
+        }
+        return true;
```
rlubos: I'd add an empty line before `return`.

src/mac_features/nrf_drv_radio802154_filter.c
```diff
@@ -0,0 +1,297 @@
...
+                                          uint8_t       * p_num_bytes,
+                                          uint8_t         frame_type,
+                                          uint8_t         frame_version)
+{
+    switch (frame_version)
+    {
+    case FRAME_VERSION_0:
+    case FRAME_VERSION_1:
+        return dst_addressing_end_offset_get_2006(p_psdu, p_num_bytes, frame_type);
+        break;
```
rlubos: No need for this `break`.

src/mac_features/nrf_drv_radio802154_filter.c
```diff
@@ -0,0 +1,297 @@
...
+
+    case FRAME_VERSION_2:
+        return dst_addressing_end_offset_get_2015(p_psdu, p_num_bytes, frame_type);
+        break;
```
rlubos: No need for this `break`.

src/mac_features/nrf_drv_radio802154_filter.c
```diff
@@ -0,0 +1,297 @@
...
+
+    default:
+        return false;
```
rlubos: Personally, I would move this `return` outside of `switch`. But if you prefer it this way, I won't argue, not sure if it's specified in FCS.

src/mac_features/nrf_drv_radio802154_filter.c
```diff
@@ -0,0 +1,297 @@
...
+    }
+}
+
+static bool dst_pan_id_check(const uint8_t *p_psdu)
```
rlubos: Missing space after `*`.

src/mac_features/nrf_drv_radio802154_filter.c
```diff
@@ -0,0 +1,297 @@
...
+{
+    if ((0 == memcmp(&p_psdu[PAN_ID_OFFSET], nrf_drv_radio802154_pib_pan_id_get(), PAN_ID_SIZE)) ||
+        (0 == memcmp(&p_psdu[PAN_ID_OFFSET], BROADCAST_ADDRESS, PAN_ID_SIZE)))
+    {
+        return true;
+    }
+
+    if (((p_psdu[FRAME_TYPE_OFFSET] & FRAME_TYPE_MASK) == FRAME_TYPE_BEACON) &&
```
rlubos: I guess that to be consistent, you should use constant first?

src/mac_features/nrf_drv_radio802154_filter.c
```diff
@@ -0,0 +1,297 @@
...
+        (0 == memcmp(nrf_drv_radio802154_pib_pan_id_get(), BROADCAST_ADDRESS, PAN_ID_SIZE)))
+    {
+        return true;
+    }
+
+    return false;
+}
+
+static bool dst_short_addr_check(const uint8_t *p_psdu)
```
rlubos: Missing space after `*`.

src/mac_features/nrf_drv_radio802154_filter.c
```diff
@@ -0,0 +1,297 @@
...
+
+    return false;
+}
+
+static bool dst_extended_addr_check(const uint8_t *p_psdu)
+{
+    if ((0 == memcmp(&p_psdu[DEST_ADDR_OFFSET],
+                     nrf_drv_radio802154_pib_extended_address_get(),
+                     EXTENDED_ADDRESS_SIZE)) ||
+        (0 == memcmp(&p_psdu[DEST_ADDR_OFFSET], BROADCAST_ADDRESS, SHORT_ADDRESS_SIZE)))
```
rlubos: Is it ok to check for broadcast address while extended address is indicated? I think that spec says no, but maybe is's specified in other place than I looked in.

src/mac_features/nrf_drv_radio802154_filter.c
```diff
@@ -0,0 +1,297 @@
...
+
+    case EXTENDED_ADDR_CHECK_OFFSET:
+        if (!dst_pan_id_check(p_psdu))
+        {
+            return false;
+        }
+
+        return dst_extended_addr_check(p_psdu);
+
+        break;
```
rlubos: Unreachable statement.

src/mac_features/nrf_drv_radio802154_filter.c
hubertmis: FCS states that "procedures may be named according to the question they answer". Not typical question-like function name.

src/mac_features/nrf_drv_radio802154_filter.c
jciupis: FCS states that all switch statements shall contain a default block. IMO all `return` statements such as in line 73 or 92 should be moved to the default block. It would unify the approach across the whole file too.

nRF-IEEE-802.15.4-radio-driver.packsc
```diff
@@ -74,6 +75,7 @@
                     "src/nrf_drv_radio802154_revision.c",
                     "src/nrf_drv_radio802154_rx_buffer.c",
                     "src/mac_features/nrf_drv_radio802154_csma_ca.c",
+                    "src/mac_features/nrf_drv_radio802154_filter.c",
```
jciupis: As the corresponding header file is included in _fsm.c, it should also be added to the `"headers"` section and to the cmock variant of the driver module.

src/mac_features/nrf_drv_radio802154_filter.c
hubertmis: On the other hand function "may have more than one exit point where it aids readability". I'll add default cases and single exit point wherever appropriate.

src/mac_features/nrf_drv_radio802154_filter.c
jciupis: It is not obvious what the purpose of this function is. Why its name is not definitive (may instead of is)?

hubertmis: This function checks if message of given type may include addressing fields. Values in Frame Control Field specify if addressing is actually present. I'll add doxygen descriptions to all internal functions. Do you think I should rename it also?

jciupis: I don't think so then - the current name seems to be a good compromise between FCS and readability. Once the documentation is added, the current name is going to be clear.

nRF-IEEE-802.15.4-radio-driver.packsc
```diff
@@ -17,7 +17,8 @@
             "src/nrf_drv_radio802154_procedures_duration.h",
             "src/nrf_drv_radio802154_request.h",
             "src/nrf_drv_radio802154_revision.h",
-            "src/nrf_drv_radio802154_rx_buffer.h"
+            "src/nrf_drv_radio802154_rx_buffer.h",
+            "src/mac_features/nrf_drv_radio802154_filter.c"
```
jciupis: I think it should be _filter.h file here. Also, unless something like `"cmock\\mock_nrf_drv_radio802154_filter.c"` is added to `"_files"` field in the cmock variant, unit test of the fsm module will not be built.

hubertmis: Sure thing!

PR #31, opened 2018-01-10
Refactor static inline functions implementation to generate mocks

jciupis:


src/nrf_drv_radio802154_procedures_duration.h
```diff
@@ -59,7 +59,21 @@
                                PHY_SHR_DURATION +                                                 \
                                (NUM_OCTETS_IN_ACK * PHY_SYMBOLS_PER_OCTET))
 
-static inline uint16_t nrf_drv_radio802154_tx_duration_get(uint8_t psdu_length,
+#ifndef __STATIC_INLINE
```
hubertmis: Is this define needed here? In nrf_radio.h there is no such define.

jciupis: nrf_radio.h includes nrf.h, which includes core_cmx.h. __STATIC_INLINE is defined there. in _procedures_duration.h there are only stdbool.h and stdint.h headers included and that is why I defined __STATIC_INLINE here.

hubertmis: Your proposal will for sure work in GCC. I'm not sure about other compilers. It would be safer to include nrf.h instead.

jciupis:
Regression test

hubertmis:
Regression test

src/nrf_drv_radio802154_procedures_duration.h
```diff
@@ -104,12 +115,14 @@ static inline uint16_t nrf_drv_radio802154_rx_duration_get(uint8_t psdu_length,
     return result;
 }
 
-static inline uint16_t nrf_drv_radio802154_cca_duration_get(void)
+__STATIC_INLINE uint16_t nrf_drv_radio802154_cca_duration_get(void)
 {
     // ramp down + rx ramp up + CCA
     uint16_t result = MAX_RAMP_DOWN_TIME + RX_RAMP_UP_TIME + A_CCA_DURATION * PHY_US_PER_SYMBOL;
```
pszkotak: nit: I see some insonsistency regarding the use of parenthesess. E.g. here vs line 110.
Maybe we should relay on the FCS and put set here as well?

hubertmis:
Regression test

hubertmis:
Regression test

PR #32, opened 2018-01-10
Add binary search algorithm for setting pending bits in ACKs

jciupis:


src/nrf_drv_radio802154_ack_pending_bit.c
```diff
@@ -60,115 +60,295 @@ static bool m_setting_pending_bit_enabled;
...
+ *
+ * @param[in]  p_first_addr     Pointer to a first address that should be compared.
+ * @param[in]  p_second_addr    Pointer to a second address that should be compared.
+ * @param[in]  extended         Indication if @p p_first_addr and @p p_second_addr are extended or short addresses.
+ *
+ * @retval -1  First address is less than the second address.
+ * @retval  0  First address is equal to the second address.
+ * @retval  1  First address is greater than the second address.
+ */
+static int8_t addr_compare(const uint8_t * p_first_addr, const uint8_t * p_second_addr, bool extended)
```
hubertmis: This function is quite long. I would split it to 3 functions: `extended_addr_compare()`, `short_addr_compare()` and `addr_compare()` that calls both other functions.

src/nrf_drv_radio802154_ack_pending_bit.c
```diff
@@ -60,115 +60,295 @@ static bool m_setting_pending_bit_enabled;
...
 {
-    uint8_t * p_empty_slot = NULL;
+    int8_t low      = 0;
+    int8_t midpoint = 0;
+    int8_t high     = extended ? (m_num_of_pending_extended) : (m_num_of_pending_short);
 
     if (extended)
     {
-        for (uint32_t i = 0; i < NUM_PENDING_EXTENDED_ADDRESSES; i++)
+        while (high >= low)
```
hubertmis: if and else blocks looks nearly the same here. You can extract it to separate function and pass pointer to right address table as argument. Of course you will need other offsets for short and extended addresses, but you can make it generic like:
```
func(uint8_t * p_addr_table, uint8_t entry_size, ...)
{
   ...
  addr_compare(p_addr, p_addr_table + entry_size * entry_index);
   ...
}

File: src/nrf_drv_radio802154_ack_pending_bit.c [hunk 3]
```diff
@@ -60,115 +60,295 @@ static bool m_setting_pending_bit_enabled;
...
+        }
+
+        for (uint32_t i = m_num_of_pending_extended; i > location; i--)
+        {
+            memcpy(m_pending_extended[i], m_pending_extended[i - 1], sizeof(m_pending_extended[i - 1]));
+        }
+
+        memcpy(m_pending_extended[location], p_addr, EXTENDED_ADDRESS_SIZE);
+        m_num_of_pending_extended++;
+        return true;
```
hubertmis: I would move `return true` out of if-else.

src/nrf_drv_radio802154_ack_pending_bit.c
```diff
@@ -60,115 +60,295 @@ static bool m_setting_pending_bit_enabled;
...
-            }
+            return false;
+        }
+
+        for (uint32_t i = location; i < m_num_of_pending_extended - 1; i++)
+        {
+            memcpy(m_pending_extended[i], m_pending_extended[i + 1], sizeof(m_pending_extended[i + 1]));
         }
+
+        memset(m_pending_extended[m_num_of_pending_extended - 1], 0, EXTENDED_ADDRESS_SIZE);
```
hubertmis: Do we have to `memset` here? We keep track of occupied elements in the array by `m_num_of_pending_extended`, not element value anymore.

src/nrf_drv_radio802154_ack_pending_bit.c
```diff
@@ -60,115 +60,320 @@ static bool m_setting_pending_bit_enabled;
...
+    int8_t  low        = 0;
+    int8_t  midpoint   = 0;
+    int8_t  high       = extended ? m_num_of_pending_extended : m_num_of_pending_short;
+    uint8_t entry_size = extended ? EXTENDED_ADDRESS_SIZE : SHORT_ADDRESS_SIZE;
+
+    while (high >= low)
+    {
+        midpoint = low + (high - low) / 2;
+
+        if (addr_compare(p_addr, p_addr_array + entry_size * midpoint, extended) == 0)
```
KaSroka: Do addr_compare once and store its result instead of calling it twice with same params

src/nrf_drv_radio802154_ack_pending_bit.c
```diff
@@ -60,115 +60,320 @@ static bool m_setting_pending_bit_enabled;
...
+ * @retval  0  First address is equal to the second address.
+ * @retval  1  First address is greater than the second address.
+ */
+static int8_t short_addr_compare(const uint8_t * p_first_addr, const uint8_t * p_second_addr)
 {
-    m_setting_pending_bit_enabled = enabled;
+    uint16_t first_addr  = *(uint16_t *)(p_first_addr);
+    uint16_t second_addr = *(uint16_t *)(p_second_addr);
+
+    if (first_addr < second_addr)
```
KaSroka: Why we are not pushing unused short addresses to the end of the list?

src/nrf_drv_radio802154_ack_pending_bit.c
```diff
@@ -60,115 +60,320 @@ static bool m_setting_pending_bit_enabled;
...
+static int8_t extended_addr_compare(const uint8_t * p_first_addr, const uint8_t * p_second_addr)
 {
-    memset(m_pending_extended, 0, sizeof(m_pending_extended));
-    memset(m_pending_short, 0xff, sizeof(m_pending_short));
-    m_setting_pending_bit_enabled = true;
+    uint64_t first_addr  = *(uint64_t *)(p_first_addr);
+    uint64_t second_addr = *(uint64_t *)(p_second_addr);
+
+    /* Push unused extended adresses to the end of the list. */
+    if ((first_addr < second_addr) || (second_addr == 0))
```
KaSroka: We will return -1 when both addresses are 0. Is it ok?

src/nrf_drv_radio802154_ack_pending_bit.c
```diff
@@ -60,115 +60,320 @@ static bool m_setting_pending_bit_enabled;
...
+        }
+        else
+        {
+            low = midpoint + 1;
+        }
+    }
+    
+    if (low == midpoint + 1)
+    {
+        midpoint++;
```
KaSroka: Why are we doing this?

src/nrf_drv_radio802154_ack_pending_bit.c
```diff
@@ -60,115 +60,320 @@ static bool m_setting_pending_bit_enabled;
...
-
-            if (0 == memcmp(m_pending_extended[i], UNUSED_PENDING_EXTENDED_ADDRESS, sizeof(m_pending_extended[i])))
-            {
-                p_empty_slot = m_pending_extended[i];
-            }
+            return false;
         }
 
-        if (p_empty_slot != NULL)
+        for (uint32_t i = m_num_of_pending_extended; i > location; i--)
```
KaSroka: This for can be replaced with single memmove()

src/nrf_drv_radio802154_ack_pending_bit.c
```diff
@@ -60,115 +60,320 @@ static bool m_setting_pending_bit_enabled;
...
-
-            if (0 == memcmp(m_pending_short[i], UNUSED_PENDING_SHORT_ADDRESS, sizeof(m_pending_short[i])))
-            {
-                p_empty_slot = m_pending_short[i];
-            }
+            return false;
         }
 
-        if (p_empty_slot != NULL)
+        for (uint32_t i = m_num_of_pending_short; i > location; i--)
```
KaSroka: Same as in 246

src/nrf_drv_radio802154_ack_pending_bit.c
```diff
@@ -60,115 +60,320 @@ static bool m_setting_pending_bit_enabled;
...
+ * @brief Add an address to the address list in ascending order.
+ *
+ * @param[in]  p_addr           Pointer to the address to be added.
+ * @param[in]  location         Index of the location where @p p_addr should be added.
+ * @param[in]  extended         Indication if @p p_addr is an extended or a short addresses.
+ *
+ * @retval true   Address @p p_addr has been added to the list successfully.
+ * @retval false  Address @p p_addr could not be added to the list.
+ */
+static bool addr_add(const uint8_t * p_addr, uint8_t location, bool extended)
```
KaSroka: I would split it to addr_add_extended and short as compare functions

src/nrf_drv_radio802154_ack_pending_bit.c
```diff
@@ -60,115 +60,320 @@ static bool m_setting_pending_bit_enabled;
...
         {
-            if (0 == memcmp(m_pending_extended[i], p_addr, sizeof(m_pending_extended[i])))
-            {
-                memset(m_pending_extended[i], 0, sizeof(m_pending_extended[i]));
-                result = true;
-            }
+            return false;
         }
+
+        for (uint32_t i = location; i < m_num_of_pending_extended - 1; i++)
```
KaSroka: Same as in 246

src/nrf_drv_radio802154_ack_pending_bit.c
```diff
@@ -60,115 +60,320 @@ static bool m_setting_pending_bit_enabled;
...
         {
-            if (0 == memcmp(m_pending_short[i], p_addr, sizeof(m_pending_short[i])))
-            {
-                memset(m_pending_short[i], 0xff, sizeof(m_pending_short[i]));
-                result = true;
-            }
+            return false;
         }
+
+        for (uint32_t i = location; i < m_num_of_pending_short - 1; i++)
```
KaSroka: Same as in 246

src/nrf_drv_radio802154_ack_pending_bit.c
```diff
@@ -60,115 +60,320 @@ static bool m_setting_pending_bit_enabled;
...
-            }
+            return false;
         }
+
+        for (uint32_t i = location; i < m_num_of_pending_extended - 1; i++)
+        {
+            memcpy(m_pending_extended[i], m_pending_extended[i + 1], sizeof(m_pending_extended[i + 1]));
+        }
+
+        m_num_of_pending_extended--;
```
KaSroka: When you are comparing extended addresses you have special case for addr set to 0 but I can't see when you set removed addr to 0.

src/nrf_drv_radio802154_ack_pending_bit.c
```diff
@@ -60,115 +60,320 @@ static bool m_setting_pending_bit_enabled;
...
+    m_num_of_pending_extended     = 0;
+    m_num_of_pending_short        = 0;
+}
+
+void nrf_drv_radio802154_ack_pending_bit_set(bool enabled)
+{
+    m_setting_pending_bit_enabled = enabled;
+}
+
+bool nrf_drv_radio802154_ack_pending_bit_for_addr_set(const uint8_t * p_addr, bool extended)
```
KaSroka: Some description of this and 344 funtions would be beneficial as now it's hard to determine what they should do.

src/nrf_drv_radio802154_ack_pending_bit.c
```diff
@@ -198,41 +403,21 @@ bool nrf_drv_radio802154_ack_pending_bit_should_be_set(const uint8_t * p_psdu)
...
-                {
-                    return true;
-                }
-            }
-
+            extended = true;
             break;
 
         default:
             return true;
```
KaSroka: Should we set pending bit by default?

KaSroka:
Also please squash fixes with base commit

src/nrf_drv_radio802154_ack_pending_bit.c
hubertmis: Yes

src/nrf_drv_radio802154_ack_pending_bit.c
hubertmis: Descriptions are in .h file

src/nrf_drv_radio802154_ack_pending_bit.c
hubertmis: I wanted to add this comment, but I verified that both versions create exactly the same assembler. I think it may stay the way it is.

hubertmis: Or do you think it will be easier to maintain to call function once and store result?

jciupis: I think it's a good idea.

KaSroka: Have you checked switch version?
```c
switch(addr_compare(...))
{
    case -1:
        ...
        break;
    case 0:
        ...
        break;
    case 1:
        ...
        break;
}
```

src/nrf_drv_radio802154_ack_pending_bit.c
jciupis: The check for addresses being equal to zero is obsolete now since all memsets have been removed. I will fix that

src/nrf_drv_radio802154_ack_pending_bit.c
KaSroka: So this one is also obsolete

src/nrf_drv_radio802154_ack_pending_bit.c
KaSroka: This one is obsolete as well

src/nrf_drv_radio802154_ack_pending_bit.c
jciupis: Yes, although here this pushing part was unnecessary anyway, because unused short address is coded 0xff and unused extended address - 0x0.

src/nrf_drv_radio802154_ack_pending_bit.c
jciupis: In general, we want to return the midpoint. However, if in the last iteration of the loop you get

list[midpoint] < addr_to_add

it means that the new element should be added after midpoint. Suppose you have a { 1, 3, 4 } list and you want to add a 2 on the second place. In the last iteration you are stuck with midpoint equal to 0 and you will get a { 2, 1, 3, 4 } list. This if() lets you use the knowledge that midpoint was smaller than the address and add it correctly.

src/nrf_drv_radio802154_ack_pending_bit.c
jciupis: I talked with @hubertmis that a generic function, such as addr_binary_search is to addr_index find, would be a solution too. What do you think would be better?

src/nrf_drv_radio802154_ack_pending_bit.c
hubertmis: No. But I believe switch-case up to 4 cases is processed as if-elseif-else

src/nrf_drv_radio802154_ack_pending_bit.c
KaSroka: I would add a comment with short description and probably this example

src/nrf_drv_radio802154_ack_pending_bit.c
jciupis: All occurences fixed.

hubertmis:
Regression test

PR #33, opened 2018-01-15
Multi-instance timer scheduler

rlubos:
Additionally added `nrf_drv_radio802154_timer_is_running` function implementation, as it was not present, yet useful in tests.


src/timer_scheduler/nrf_drv_radio802154_timer_sched.c
```diff
@@ -66,39 +184,185 @@ void nrf_drv_radio802154_timer_sched_add(nrf_drv_radio802154_timer_t * p_timer,
...
+    while (true)
+    {
+        queue_cntr = m_queue_changed_cntr;
+        pp_item    = (nrf_drv_radio802154_timer_t **)&mp_head;
+        p_next     = NULL;
+
+        // Search the current queue to find appropriate position to insert timer.
+        while (true)
+        {
+            nrf_drv_radio802154_timer_t * p_cur = (nrf_drv_radio802154_timer_t *)__LDREXW((uint32_t *)pp_item);
```
hubertmis: Do we need to cast `pp_item` to `uint32_t *` here?

src/timer_scheduler/nrf_drv_radio802154_timer_sched.c
```diff
@@ -66,39 +184,185 @@ void nrf_drv_radio802154_timer_sched_add(nrf_drv_radio802154_timer_t * p_timer,
...
+        if (queue_cntr != m_queue_changed_cntr)
+        {
+            // Higher priority modified the queue while iterating, try again.
+            continue;
+        }
 
-    nrf_drv_radio802154_timer_start(p_timer->t0, dt);
+        p_timer->p_next = p_next;
+
+        if (!__STREXW((uint32_t)p_timer, (uint32_t *)pp_item))
```
hubertmis: Let's say there is a queue: A->B->D. We are going to put C between B and D from low priority context. During `add` operation between `if (queue_cntr != m_queue_changed_cntr)` and `__STREXW` higher priority removes B. Are we safe in this case? Are we going to add C between A and D?
Edit: I've found the answer in `remove` function.

src/timer_scheduler/nrf_drv_radio802154_timer_sched.c
```diff
@@ -66,39 +184,185 @@ void nrf_drv_radio802154_timer_sched_add(nrf_drv_radio802154_timer_t * p_timer,
...
 
-    mp_timer = NULL;
+    // Write to the pointer next on removal to ensure that node removal is detected by
+    // lower pritority context in case it was going to be used.
+    if (p_cur != NULL)
+    {
+        do
+        {
+            (void)__LDREXW((uint32_t *)&p_cur->p_next);
+        } while (__STREXW(0, (uint32_t *)&p_cur->p_next));
```
hubertmis: Oh, I can see answer to question from line 234. I would use `NULL` instead of `0` when storing pointer.

src/timer_scheduler/nrf_drv_radio802154_timer_sched.c
```diff
@@ -66,39 +184,185 @@ void nrf_drv_radio802154_timer_sched_add(nrf_drv_radio802154_timer_t * p_timer,
...
-    return p_timer == mp_timer;
+    uint8_t queue_cntr;
+    bool    result;
+
+    do
+    {
+        result     = false;
+        queue_cntr = m_queue_changed_cntr;
+
+        for (volatile nrf_drv_radio802154_timer_t  * p_cur = mp_head; p_cur != NULL; p_cur++)
```
hubertmis: `p_cur++` is correct here? I think we should use `p_cur = p_cur->p_next`.

src/timer_scheduler/nrf_drv_radio802154_timer_sched.c
```diff
@@ -66,39 +184,185 @@ void nrf_drv_radio802154_timer_sched_add(nrf_drv_radio802154_timer_t * p_timer,
...
+    } while (queue_cntr != m_queue_changed_cntr);
+
+    return result;
 }
 
 void nrf_drv_radio802154_timer_fired(void)
 {
-    assert(mp_timer != NULL);
-    assert(mp_timer->callback != NULL);
+    assert(mp_head != NULL);
```
hubertmis: I'm afraid it may happen that during this function higher priority context removes head. Maybe we could copy pointer to head and to callback to avoid hardfault?

src/timer_scheduler/nrf_drv_radio802154_timer_sched.c
```diff
@@ -44,9 +46,125 @@
...
+
+/** @brief Increment queue counter value to detect changes in the queue. */
+static inline void queue_cntr_bump(void)
+{
+    volatile uint8_t cntr;
+
+    do
+    {
+        cntr = __LDREXB(&m_queue_changed_cntr);
+    } while (__STREXB(++cntr, &m_queue_changed_cntr));
```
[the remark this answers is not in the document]
rlubos: No reason, must've been easier to type. But is there a diffrence with optimization?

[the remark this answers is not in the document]
hubertmis: I think there is no difference. Just I didn't know if you intentionally increment `cntr` variable (is it needed). I'm OK to leave it as it is if you prefer this way.

src/timer_scheduler/nrf_drv_radio802154_timer_sched.c
```diff
@@ -66,39 +184,185 @@ void nrf_drv_radio802154_timer_sched_add(nrf_drv_radio802154_timer_t * p_timer,
     assert(p_timer != NULL);
     assert(p_timer->callback != NULL);
 
-    uint32_t dt = p_timer->dt;
-
     if (round_up)
     {
-        dt += nrf_drv_radio802154_timer_granularity_get() - 1;
+        p_timer->dt += nrf_drv_radio802154_timer_granularity_get() - 1;
```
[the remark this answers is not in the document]
rlubos: We have to unless we store information about rounding up/down. There is a note about rounding up already, I'll add a word or two that `dt` is modified.

src/timer_scheduler/nrf_drv_radio802154_timer_sched.c
rlubos: AFAIR compiler complained.

src/timer_scheduler/nrf_drv_radio802154_timer_sched.c
rlubos: That's what I've originally wanted to do, but compiler expected to typecast to `uint32_t`. I find passing `0` neater than `(uint32_t)NULL`.

hubertmis: I prefer `(uint32_t)NULL`, but I'm OK to leave it.

src/timer_scheduler/nrf_drv_radio802154_timer_sched.c
rlubos: Right! Good catch. I had my timers arranged in an array during tests so it must've slipped through.

src/timer_scheduler/nrf_drv_radio802154_timer_sched.c
rlubos: Right, will update :+1:

hubertmis:
@pszkotak Any comments on this one?

hubertmis:
@rlubos, rebase please. I'll merge shortly after.

rlubos:
Done

PR #35, opened 2018-01-17
Fix mock generation for unit tests.

hubertmis:


nRF-IEEE-802.15.4-radio-driver.packsc
```diff
@@ -102,6 +102,7 @@
                     "cmock\\mock_nrf_drv_radio802154_ack_pending_bit.c",
                     "cmock\\mock_nrf_drv_radio802154_critical_section.c",
                     "cmock\\mock_nrf_drv_radio802154_debug.c",
+                    "cmock\\mock_nrf_drv_radio802154_filter.c",
```
jciupis: Great! Please also add `"src/mac_features"` to `"_includes"` section of the nrf_drv_radio802154 module. The subfolder has has to be included explicitly, so that _filter.h header can be recognized when compiling the mock file.

hubertmis: Fixed

PR #36, opened 2018-01-18
Fix IAR build of timer scheduler

rlubos:


src/timer_scheduler/nrf_drv_radio802154_timer_sched.c
```diff
@@ -144,7 +144,14 @@ static inline void handle_timer(void)
             }
             else
             {
-                nrf_drv_radio802154_timer_start(p_head->t0, p_head->dt);
+                uint32_t t0 = p_head->t0;
+                uint32_t dt = p_head->dt;
+
+                // Set the timer only if current HEAD wasn't removed - otherwise t0 and dt might not be correct.
+                if (p_head == mp_head)
```
hubertmis: What happens if head is removed between this `if ()` and call to `_timer_start()`?

rlubos: We've already stored valid `t0` and `dt`, so we won't set garbage to the timer. On the other hand, as head was changed, the exterior loop will reiterate and we either set the timer to updated head, or stop the timer.

hubertmis: Is this `if ()` needed then? As you said outer loop shall iterate anyway.

rlubos: Well, my intention was to protect against situation, when higher priority interrupts us between reading `t0` and `dt`, and modifies `dt`. There is a chance that we get a combination of unrelated 't0' and 'dt' that would indicate a time in the past, so in result we would erroneously call the timer handler, which is a problem in current implementation.

hubertmis: OK. Could you add it to the comment that it protects from getting t0 and dt from other timers?

PR #39, opened 2018-01-24
Fix IAR build of the driver with CSMA-CA disabled

jciupis:


src/nrf_drv_radio802154_fsm_hooks.c
```diff
@@ -54,20 +54,26 @@ static const abort_hook m_abort_hooks[] =
 {
 #if NRF_DRV_RADIO802154_CSMA_CA_ENABLED
     nrf_drv_radio802154_csma_ca_abort,
+#else
+    NULL,
```
hubertmis: Is it OK? I'm afraid we will end in HardFault if function below will try to branch to NULL pointer.

jciupis: I have assumed by mistake that in this solution the for loop will not be executed, but it would be executed once, leading to HardFault. You are right - I will commit a proper fix then.

KaSroka:
So IAR doesn't like empty arrays?

src/nrf_drv_radio802154_fsm_hooks.c
```diff
@@ -50,31 +50,50 @@ typedef void (* abort_hook)(void);
 typedef bool (* tx_failed_hook)(nrf_drv_radio802154_tx_error_t error);
 typedef bool (* tx_started_hook)(void);
 
+/* Since some compilers do not allow empty initializers for arrays with unspecified bounds, 
+ * NULL pointer is appended to below arrays if the compiler used is not GCC. It is never reached
```
KaSroka: In current implementation it is reached.

PR #40, opened 2018-01-24
Add ACK timeout feature

KaSroka:


src/mac_features/nrf_drv_radio802154_ack_timeout.c
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * @file
+ *   This file implements ACK timeout procedure for the 802.15.4 driver.
+ *
+ */
+
+#include "nrf_drv_radio802154_ack_timeout.h"
+
+#include <assert.h>
```
hubertmis: Is assert or stdlib used by this file?

src/mac_features/nrf_drv_radio802154_ack_timeout.h
```diff
@@ -0,0 +1,83 @@
...
+ * @defgroup nrf_driver_radio802154_csma_ca 802.15.4 driver ACK timeout support
+ * @{
+ * @ingroup nrf_driver_radio802154
+ * @brief ACK timeout feature.
+ */
+
+/**
+ * @brief Set timeout time for ACK timeout feature.
+ * 
+ * @param[in]  time  Timeout time in us.
```
hubertmis: Please, add note that timeout is started at beginning of frame transmission (after transmission of PHR) and that default value is in config.h.

src/nrf_drv_radio802154_config.h
```diff
@@ -322,6 +322,29 @@ extern "C" {
 #define NRF_DRV_RADIO802154_CSMA_CA_MAX_CSMA_BACKOFFS 4
 #endif
 
+/*******************************************************************************
+ * @section ACK timeout feature configuration.
+ ******************************************************************************/
+
+/**
+ * @def NRF_DRV_RADIO802154_CSMA_CA_ENABLED
```
jciupis: _ACK_TIMEOUT instead of _CSMA_CA? Same in 340

src/mac_features/nrf_drv_radio802154_ack_timeout.c
```diff
@@ -0,0 +1,96 @@
...
+#include "timer_scheduler/nrf_drv_radio802154_timer_sched.h"
+
+static uint32_t                    m_timeout = NRF_DRV_RADIO802154_ACK_TIMEOUT_DEFAULT_TIMEOUT; ///< ACK timeout in us.
+static nrf_drv_radio802154_timer_t m_timer;                                                     ///< Timer used to notify when we are waiting too long for ACK.
+
+static void timeout_timer_fired(void * p_context)
+{
+    (void)p_context;
+
+    nrf_drv_radio802154_request_receive();
```
hubertmis: You should handle case when this function returns `false`

KaSroka: I was thinking about this case. Shouldn't we notify about tx failed even when we cannot switch to receive?

hubertmis: We should switch to receive if we notify. Perhaps we could restart timer with shorter delay in case of `false` here?

PR #43, opened 2018-01-29
Refactored energy detection procedure

hubertmis:


src/nrf_drv_radio802154_fsm.c
```diff
@@ -2837,10 +2941,23 @@ bool nrf_drv_radio802154_fsm_transmit(nrf_drv_radio802154_term_t term_lvl,
...
 
-    bool result = false;
+    if (result)
+    {
+        state_set(RADIO_STATE_ED);
+        m_ed_time_left = time_us;
+        m_ed_result    = 0;
+    }
+
+    if (result)
```
rlubos: `result` does not seem to be modified inbetween, why separate `if (result)`?

src/nrf_drv_radio802154_fsm.c
```diff
@@ -2837,10 +2941,23 @@ bool nrf_drv_radio802154_fsm_transmit(nrf_drv_radio802154_term_t term_lvl,
...
+    {
+        ed_begin(true);
+    }
+
+    return result;
+
+    result = false;
```
rlubos: It doesn't look like reachable code. Shouldn't it be removed?

hubertmis: Currently there is a lot of unreachable code. I'm going to remove it at the end or refactoring. Right now I want to keep it as reference.

rlubos: I understand that, but how can reviewer tell if something is left by author on purpose or if it is a bug? I would wrap it inside `#if 0 ... #endif`, so there will be no misunderstandigs (like you did before by the way).

src/nrf_drv_radio802154_fsm.c
hubertmis: It will be probably modified by RAAL functions.

PR #45, opened 2018-01-30
Standalone CCA implementation + fix for bug with not cleared PPI fork

hubertmis:


src/nrf_drv_radio802154_fsm.c
```diff
@@ -1357,6 +1380,71 @@ static void ed_begin(bool disabled_was_triggered)
     }
 }
 
+static void cca_begin(bool disabled_was_triggered)
+{
+    bool trigger_disable;
```
jciupis: I would consider renaming this variable to "trigger_task_disable" or similar, now it could be mistaken for disabling a trigger.

src/nrf_drv_radio802154_fsm.c
```diff
@@ -1329,34 +1353,98 @@ static void ed_begin(bool disabled_was_triggered)
...
+    {
+        trigger_task_disable = false;
+    }
+
+    if (trigger_task_disable)
+    {
+        nrf_radio_task_trigger(NRF_RADIO_TASK_DISABLE);
+    }
+
+    if (!nrf_raal_timeslot_request(nrf_drv_radio802154_cca_duration_get()))
```
rlubos: Is it intended to request for a timeslot twice?

hubertmis: It is. It was solving possible problem with Timeslot API. Probably need to rethink it, this solution is not the best. I'll merge it as it is now and update it during RAAL implementation in TX path (there is the same problem).

hubertmis:
Closing this PR and reopening from another branch. I have too much commits here.

PR #46, opened 2018-01-30
Add frame filtering to rx path in fsm

rlubos:


src/nrf_drv_radio802154_fsm.c
```diff
@@ -115,7 +123,8 @@ static inline uint32_t short_phyend_disable_mask_get(void)
                                short_phyend_disable_mask_get())
 
 #define SHORTS_RX_ACK         (NRF_RADIO_SHORT_ADDRESS_RSSISTART_MASK |                            \
-                               NRF_RADIO_SHORT_END_DISABLE_MASK)
+                               NRF_RADIO_SHORT_END_DISABLE_MASK |                                  \
+                               SHORT_ADDRESS_BCSTART)
```
hubertmis: Do we use BCC for rx Ack?

src/nrf_drv_radio802154_fsm.c
```diff
@@ -1490,77 +1513,52 @@ static inline void irq_bcmatch_state_rx_header(void)
...
+                }
+                else
+                {
+                    // Promiscuous mode, allow incorrect frames. Nothing to do here.
                 }
             }
 
             break;
 
         case NRF_RADIO_STATE_TX_IDLE:
```
hubertmis: We should not handle this case here anymore. It will be handled by CRCERROR or CRCOK handlers.

src/nrf_drv_radio802154_fsm.c
```diff
@@ -1572,133 +1570,181 @@ static inline void irq_bcmatch_state_rx_header(void)
 static inline void irq_crcok_state_rx(void)
```
hubertmis: It seems that CRCERROR must be handled as well as BCMATCH. CRCERROR automaticly restarts rx procedure. Event handler should reset BCC register and clear frame_filtered flag.

src/nrf_drv_radio802154_fsm.c
rlubos: Do you think it is not needed? My intention was to use it, but I can see now that some code is missing in `PHYEND` handler.

hubertmis: I think RX_ACK needs another filtering. It should use MHRMATCH feature of the RADIO peripheral. There is separated task to implement it.

rlubos:
I've applied fixes in a separate commit for clarity, will squash before merge.

src/nrf_drv_radio802154_fsm.c
```diff
@@ -1154,6 +1154,8 @@ static void receive_begin(bool disabled_was_triggered)
 #if !NRF_DRV_RADIO802154_DISABLE_BCC_MATCHING
     nrf_radio_event_clear(NRF_RADIO_EVENT_BCMATCH);
     nrf_radio_int_enable(NRF_RADIO_INT_BCMATCH_MASK);
+    nrf_radio_event_clear(NRF_RADIO_EVENT_CRCERROR);
```
hubertmis: Nit: nrf_radio_event_clear, nrf_radio_int_enable, _disable may be called once with arguments concatenated with `|`

PR #47, opened 2018-01-31
FEM support for refactored FSM RX path

jciupis:


src/nrf_drv_radio802154_fsm.c
```diff
@@ -365,17 +365,15 @@ static void cca_configuration_update(void)
 /** Trigger RX task. */
 static void rx_enable(void)
```
hubertmis: This function is unused after FSM refactoring.

src/nrf_drv_radio802154_fsm.c
```diff
@@ -365,17 +365,15 @@ static void cca_configuration_update(void)
...
 {
-    nrf_fem_control_time_latch();
     nrf_radio_task_trigger(NRF_RADIO_TASK_RXEN);
-    nrf_fem_control_lna_set(false);
+    nrf_fem_control_lna_set();
 }
 
 /** Trigger TX task. */
 static void tx_enable(void)
```
hubertmis: This function is also unused. Will be removed during clean-up.

src/nrf_drv_radio802154_fsm.c
```diff
@@ -860,6 +861,9 @@ static void tx_ack_terminate(void)
                           NRF_RADIO_INT_PHYEND_MASK : NRF_RADIO_INT_END_MASK);
     nrf_radio_shorts_set(SHORTS_IDLE);
     nrf_radio_task_trigger(NRF_RADIO_TASK_DISABLE);
+
+    nrf_fem_control_lna_clr();
```
hubertmis: I don't think we should explicitly clear FEM pins in `_terminate()` functions. Each `_terminate()` function triggers `DISABLE` task and FEM should clear both pins on `DISABLED` radio event.

src/nrf_drv_radio802154_fsm.c
```diff
@@ -1136,12 +1148,15 @@ static void receive_begin(bool disabled_was_triggered)
...
     // TODO: Set FEM
+    nrf_fem_control_lna_set();
+    lna_target_time = nrf_fem_control_lna_timeout_calc();
+    pa_target_time  = nrf_fem_control_pa_timeout_calc(false);
 
     // Set TIMERs
     nrf_timer_shorts_enable(NRF_DRV_RADIO802154_TIMER_INSTANCE, NRF_TIMER_SHORT_COMPARE0_STOP_MASK | NRF_TIMER_SHORT_COMPARE2_STOP_MASK);
-    nrf_timer_cc_write(NRF_DRV_RADIO802154_TIMER_INSTANCE, NRF_TIMER_CC_CHANNEL0, 1);
+    nrf_timer_cc_write(NRF_DRV_RADIO802154_TIMER_INSTANCE, NRF_TIMER_CC_CHANNEL0, lna_target_time);
     nrf_timer_cc_write(NRF_DRV_RADIO802154_TIMER_INSTANCE, NRF_TIMER_CC_CHANNEL1, 1 + 192 - 40 -23); // 40 is ramp up, 23 is event latency
```
hubertmis: `1` in this and next formula should be replaced with `lna_target_time`

src/fem/nrf_fem_control_api.h
```diff
@@ -173,6 +184,10 @@ void nrf_fem_control_lna_set(bool shorts_used);
 #define nrf_fem_control_time_latch(...)
 #define nrf_fem_control_pa_set(...)
 #define nrf_fem_control_lna_set(...)
+#define nrf_fem_control_pa_clr(...)
+#define nrf_fem_control_lna_clr(...)
+#define nrf_fem_control_pa_timeout_calc(...)
+#define nrf_fem_control_lna_timeot_calc(...)
```
hubertmis: Let's define it to `1` to return any value if FEM is disabled. It would set timer to 1 us. It should be fine.

src/fem/nrf_fem_control_api.h
```diff
@@ -173,6 +184,10 @@ void nrf_fem_control_lna_set(bool shorts_used);
 #define nrf_fem_control_time_latch(...)
 #define nrf_fem_control_pa_set(...)
 #define nrf_fem_control_lna_set(...)
+#define nrf_fem_control_pa_clr(...)
+#define nrf_fem_control_lna_clr(...)
+#define nrf_fem_control_pa_timeout_calc(...)
```
hubertmis: Let's define it to `1`. Or `0`. I didn't think it through if `0` makes any sense here :)

src/fem/nrf_fem_control_api.h
```diff
@@ -125,44 +125,55 @@ void nrf_fem_control_cfg_get(nrf_fem_control_cfg_t * p_cfg);
...
+ * This function disables a PPI that sets the pin.
+ *
+ * */
+void nrf_fem_control_lna_clr(void);
+
+/**@brief Calculate target time for a timer, which activates the Power Amplifier pin.
+ * 
+ * @param[in] turnaround Information if TX mode is entered from RX mode (turnaround) or DISABLED (ramp up).
+ *
+ * This function disables a PPI that sets the pin.
```
hubertmis: Why does this function disable PPI?

src/fem/nrf_fem_control_common.c
```diff
@@ -150,75 +203,90 @@ void nrf_fem_control_activate(void)
...
+    uint32_t target_time = 1;
+
+    if (m_nrf_fem_control_cfg.pa_cfg.enable)
     {
-        m_time_latch = nrf_fem_control_timer_time_get();
+        /* Calculate timer target time. */
+        uint32_t latency = turnaround ? NRF_FEM_RADIO_TX_TURNAROUND_LATENCY_US :
+                                        NRF_FEM_RADIO_TX_STARTUP_LATENCY_US;
+
+        uint32_t tifs = nrf_radio_ifs_get();
```
hubertmis: We don't use IFS anymore after refactoring. Remove it from this function assuming it is `0`.

src/fem/nrf_fem_control_common.c
```diff
@@ -150,75 +203,90 @@ void nrf_fem_control_activate(void)
...
+        
+        target_time = tifs <= latency ? latency - NRF_FEM_PA_TIME_IN_ADVANCE :
+                                        tifs - NRF_FEM_RADIO_TIFS_DRIFT_US - NRF_FEM_PA_TIME_IN_ADVANCE;
+
+        target_time -= nrf_fem_control_irq_delay_get();
```
hubertmis: In current approach there is no IRQ delay. also it seems that this function is removed.

src/fem/nrf_fem_control_common.c
```diff
@@ -150,75 +203,90 @@ void nrf_fem_control_activate(void)
...
+    if (m_nrf_fem_control_cfg.lna_cfg.enable)
     {
-        gpiote_configure(m_nrf_fem_control_cfg.pa_cfg.gpio_pin,
-                         (nrf_gpiote_outinit_t)!m_nrf_fem_control_cfg.pa_cfg.active_high);
-        ppi_enable();
-
-        uint32_t tifs = nrf_radio_ifs_get();
+        /* Calculate timer target time. */
+        target_time = NRF_FEM_RADIO_RX_STARTUP_LATENCY_US -
+                      NRF_FEM_LNA_TIME_IN_ADVANC - nrf_fem_control_irq_delay_get();
```
hubertmis: It looks that `_irq_delay_get()` function is removed.

src/fem/nrf_fem_control_common.c
```diff
@@ -88,39 +105,80 @@ __STATIC_INLINE void gpiote_configure(uint8_t pin, nrf_gpiote_outinit_t init_val
...
+                (uint32_t)(&NRF_GPIOTE->TASKS_SET[m_nrf_fem_control_cfg.pa_gpiote_ch_id]));
+        }
+    }
 }
 
 /** Enable PPI. */
 __STATIC_INLINE void ppi_enable(void)
 {
-    NRF_PPI->CHENSET = (1 << m_nrf_fem_control_cfg.ppi_ch_id_set) |
+    NRF_PPI->CHENSET = (1 << m_nrf_fem_control_cfg.ppi_ch_id_pa_set)  |
```
hubertmis: Why do you set registers directly instead of using hardware access layer?

src/fem/nrf_fem_control_common.c
```diff
@@ -150,75 +203,90 @@ void nrf_fem_control_activate(void)
...
-
     if (m_nrf_fem_control_cfg.lna_cfg.enable)
     {
-        gpiote_configure(m_nrf_fem_control_cfg.lna_cfg.gpio_pin,
+        /* GPIOTE configuration for TIMER->COMPARE[0] --> set LNA pin PPI */
+        gpiote_set_task_configure(m_nrf_fem_control_cfg.lna_cfg.gpio_pin,
+                         m_nrf_fem_control_cfg.lna_gpiote_ch_id,
+                         (bool)m_nrf_fem_control_cfg.lna_cfg.active_high,
                          (nrf_gpiote_outinit_t)!m_nrf_fem_control_cfg.lna_cfg.active_high);
         ppi_enable();
```
hubertmis: Why all PPIs are enabled here? Instead of only the PPI that sets LNA pin?

src/fem/nrf_fem_control_common.c
```diff
@@ -150,75 +203,90 @@ void nrf_fem_control_activate(void)
...
-void nrf_fem_control_lna_set(bool shorts_used)
+void nrf_fem_control_lna_set(void)
 {
-    uint32_t target_time;
-
     if (m_nrf_fem_control_cfg.lna_cfg.enable)
     {
-        gpiote_configure(m_nrf_fem_control_cfg.lna_cfg.gpio_pin,
+        /* GPIOTE configuration for TIMER->COMPARE[0] --> set LNA pin PPI */
+        gpiote_set_task_configure(m_nrf_fem_control_cfg.lna_cfg.gpio_pin,
```
hubertmis: I would configure GPIOTE on module initialization and/or on configuration setup. Is there any profit with re-configuring it each time PPIs are configured?

src/nrf_drv_radio802154_fsm.c
```diff
@@ -1136,12 +1148,15 @@ static void receive_begin(bool disabled_was_triggered)
     nrf_radio_int_enable(NRF_RADIO_INT_CRCOK_MASK);
 
     // TODO: Set FEM
```
hubertmis: `TODO` seems not valid anymore

src/nrf_drv_radio802154_fsm.c
```diff
@@ -1592,6 +1609,8 @@ static inline void irq_crcok_state_rx(void)
         // Set PPIs
         nrf_ppi_channel_endpoint_setup(PPI_CH2, (uint32_t) nrf_timer_event_address_get(NRF_DRV_RADIO802154_TIMER_INSTANCE, NRF_TIMER_EVENT_COMPARE1), (uint32_t) nrf_radio_task_address_get(NRF_RADIO_TASK_TXEN));
         // TODO: Set FEM PPIs
+        nrf_fem_control_lna_clr();
```
hubertmis: I don't think we should clear LNA here. It should be cleared by PPIs on `DISABLED` event.

src/fem/nrf_fem_control_api.h
```diff
@@ -125,44 +125,55 @@ void nrf_fem_control_cfg_get(nrf_fem_control_cfg_t * p_cfg);
...
 
-/**@brief Activate Power Amplifier (TX) pin of the Front End Module.
+/**@brief Activate Low Noise Amplifier (RX) pin of the Front End Module.
+ *
+ * This function will also set up a PPI to deactivate the pin on radio DISABLED event.
  *
- * @param[in] shorts_used Information if this operation is related to a task triggered by a short.
- * @param[in] turnaround  Information if TX mode is entered from RX mode (turnaround) or DISABLED (ramp up).
+ * */
+void nrf_fem_control_lna_set(void);
```
hubertmis: It seems that description of this function is invalid. Also I don't like the name. This function should configure PPI to set LNA pin on timer event. I would call it `nrf_fem_lna_ppi_enable` or something like this.
There should be also a function like `nrf_fem_lna_ppi_disable` that cleans up PPI setup when procedure that uses LNA exits.

src/nrf_drv_radio802154_fsm.c
hubertmis: Oh, I see you modified it to compile it. You could have also `#if 0` it :)

src/nrf_drv_radio802154_fsm.c
hubertmis: On the other hand, each `_terminate()` function should disable FEM PPIs that were enabled for operation being terminated. The missing function I mentioned above should be used for this purpose.

src/nrf_drv_radio802154_fsm.c
```diff
@@ -1250,6 +1265,8 @@ static void transmit_begin(const uint8_t * p_data, bool cca, bool disabled_was_t
     }
 
     // TODO: Set FEM
+    nrf_fem_control_pa_clr();
```
hubertmis: Let's remove it from this review. We will focus on TX path in another PR.

src/nrf_drv_radio802154_fsm.c
hubertmis: But LNA PPIs should be disabled by the missing function I already mentioned.

src/nrf_drv_radio802154_fsm.c
```diff
@@ -1592,6 +1609,8 @@ static inline void irq_crcok_state_rx(void)
         // Set PPIs
         nrf_ppi_channel_endpoint_setup(PPI_CH2, (uint32_t) nrf_timer_event_address_get(NRF_DRV_RADIO802154_TIMER_INSTANCE, NRF_TIMER_EVENT_COMPARE1), (uint32_t) nrf_radio_task_address_get(NRF_RADIO_TASK_TXEN));
         // TODO: Set FEM PPIs
+        nrf_fem_control_lna_clr();
+        nrf_fem_control_pa_set();
```
hubertmis: This PPI that is enabled here should be disabled in PHYEND event handler in TX_ACK state.

src/fem/nrf_fem_control_common.c
```diff
@@ -150,75 +203,90 @@ void nrf_fem_control_activate(void)
...
+    if (m_nrf_fem_control_cfg.pa_cfg.enable)
     {
-        m_time_latch = nrf_fem_control_timer_time_get();
+        /* Calculate timer target time. */
+        uint32_t latency = turnaround ? NRF_FEM_RADIO_TX_TURNAROUND_LATENCY_US :
+                                        NRF_FEM_RADIO_TX_STARTUP_LATENCY_US;
+
+        uint32_t tifs = nrf_radio_ifs_get();
+        
+        target_time = tifs <= latency ? latency - NRF_FEM_PA_TIME_IN_ADVANCE :
```
hubertmis: I can't find where `NRF_FEM_PA_TIME_IN_ADVANCE` is defined. Its value has changed since last implementation. It should be 23us instead of 5us now.

src/nrf_drv_radio802154_fsm.c
hubertmis: Or instead of enabling / disabling PPIs you can reconfigure single PPI to use 1 instead of 2. Would it be possible?

src/fem/nrf_fem_control_common.c
```diff
@@ -150,75 +203,90 @@ void nrf_fem_control_activate(void)
...
+{
+    nrf_gpiote_task_disable(m_nrf_fem_control_cfg.pa_gpiote_ch_id);
+    NRF_PPI->CHENCLR = (1 << m_nrf_fem_control_cfg.ppi_ch_id_pa_set);
+}
+
+void nrf_fem_control_lna_clr(void)
+{
+    nrf_gpiote_task_disable(m_nrf_fem_control_cfg.lna_gpiote_ch_id);
+    NRF_PPI->CHENCLR = (1 << m_nrf_fem_control_cfg.ppi_ch_id_lna_set);
+}
```
rlubos: No new line at the end of the file, IAR will complain.

src/fem/nrf_fem_control_common.c
rlubos: AFAIR `nrf_fem_control_config.h`

src/nrf_drv_radio802154_fsm.c
rlubos: Isn't this exactly what this function does? I agree that the naming is a little bit confusing though...

src/nrf_drv_radio802154_fsm.c
rlubos: Originally we've used a single PPI to enable LNA/PA. Not sure if it is still valid for this short-orientated implementation, will have to investigate.

src/fem/nrf_fem_control_common.c
```diff
@@ -88,39 +105,80 @@ __STATIC_INLINE void gpiote_configure(uint8_t pin, nrf_gpiote_outinit_t init_val
...
+        (uint32_t)(&NRF_GPIOTE->TASKS_OUT[m_nrf_fem_control_cfg.lna_gpiote_ch_id]));
+
+    /* TIMER2->COMPARE --> set PA PPI */
+    nrf_ppi_channel_endpoint_setup(
+        (nrf_ppi_channel_t)m_nrf_fem_control_cfg.ppi_ch_id_pa_set,
+        (uint32_t)(&NRF_FEM_TIMER_INSTANCE->EVENTS_COMPARE[NRF_FEM_TIMER_PA_CH]),
+        (uint32_t)(&NRF_GPIOTE->TASKS_OUT[m_nrf_fem_control_cfg.pa_gpiote_ch_id]));
+
+    /* RADIO_DISABLED --> clr LNA & clr PA PPI */
+    if (m_nrf_fem_control_cfg.lna_cfg.active_high)
```
rlubos: This condition seems a little bulky for me, have you considered using a single `nrf_ppi_channel_and_fork_endpoint_setup` call and using `m_nrf_fem_control_cfg.lna_cfg.active_high ? CLR : SET` construct instead? Thoughts?

src/fem/nrf_fem_control_api.h
```diff
@@ -125,44 +125,55 @@ void nrf_fem_control_cfg_get(nrf_fem_control_cfg_t * p_cfg);
...
+
+/**@brief Calculate target time for a timer, which activates the Power Amplifier pin.
+ * 
+ * @param[in] turnaround Information if TX mode is entered from RX mode (turnaround) or DISABLED (ramp up).
+ *
+ * This function disables a PPI that sets the pin.
  *
- * @param[in] shorts_used Information if this operation is related to a task triggered by a short.
+ * */
+uint32_t nrf_fem_control_pa_timeout_calc(bool turnaround);
```
rlubos: I think that `timeout` in this case a little bit confusing term, how about `nrf_fem_control_pa_enable_time_get` or something like this? Same would apply to LNA.

src/nrf_drv_radio802154_fsm.c
hubertmis: I think not in all cases (probably in CCA_TX we will need 3-5 PPIs for FEM), but in RX path single PPI should be enough.

src/fem/nrf_fem_control_api.h
hubertmis: Of `_delay_get`?

src/fem/nrf_fem_control_common.c
```diff
@@ -38,15 +38,17 @@
...
 #include "compiler_abstraction.h"
 #include "nrf_fem_control_config.h"
-#include "nrf_fem_control_internal.h"
 #include "nrf.h"
 #include "hal/nrf_gpio.h"
 #include "hal/nrf_gpiote.h"
 #include "hal/nrf_ppi.h"
 #include "hal/nrf_radio.h"
 
+#define NRF_FEM_TIMER_INSTANCE         NRF_TIMER1
```
hubertmis: You should use `NRF_DRV_RADIO802154_TIMER_INSTANCE` here from nrf_drv_radio802154_config.h

src/fem/nrf_fem_control_common.c
```diff
@@ -150,75 +162,93 @@ void nrf_fem_control_activate(void)
...
 
-void nrf_fem_control_time_latch(void)
+void nrf_fem_control_pa_ppi_enable(void)
 {
-    if (m_nrf_fem_control_cfg.pa_cfg.enable || m_nrf_fem_control_cfg.lna_cfg.enable)
+    if (m_nrf_fem_control_cfg.pa_cfg.enable)
     {
-        m_time_latch = nrf_fem_control_timer_time_get();
+        /* GPIOTE configuration for TIMER->COMPARE[0] --> set LNA pin PPI */
+        gpiote_task_configure(m_nrf_fem_control_cfg.lna_cfg.gpio_pin,
```
hubertmis: Is it necessary to reconfigure GPIOTE each time PPI is configured? It should be enough to configure it on initialization / configuration update.

src/fem/nrf_fem_control_common.c
```diff
@@ -150,75 +162,93 @@ void nrf_fem_control_activate(void)
...
-        nrf_fem_control_timer_set(target_time);
-    }
+void nrf_fem_control_lna_ppi_disable(void)
+{
+    nrf_gpiote_task_disable(m_nrf_fem_control_cfg.lna_gpiote_ch_id);
+    nrf_ppi_channel_disable((nrf_ppi_channel_t)m_nrf_fem_control_cfg.ppi_ch_id_set);
 }
 
-void nrf_fem_control_lna_set(bool shorts_used)
+uint32_t nrf_fem_control_pa_delay_get(bool turnaround)
```
hubertmis: It seems that `turnaround` is not needed anymore. We should use only STARTUP latency. In case of turnaround, PA should be triggered at the time turnaround starts.

src/nrf_drv_radio802154_fsm.c
```diff
@@ -753,6 +736,9 @@ static void rx_terminate(void)
     nrf_radio_int_disable(NRF_RADIO_INT_CRCOK_MASK);
     nrf_radio_shorts_set(SHORTS_IDLE);
     nrf_radio_task_trigger(NRF_RADIO_TASK_DISABLE);
+
+    nrf_fem_control_lna_ppi_disable();
```
hubertmis: Is it necessary to disable both LNA na PA PPIs in each termination function? I think in RX state only LNA is configured, while in PA state only PA is configured.

src/fem/nrf_fem_control_common.c
jciupis: You are right, having separate GPIOTE channels for PA and LNA pins makes it unnecessary.

src/fem/nrf_fem_control_api.h
```diff
@@ -125,54 +121,62 @@ void nrf_fem_control_cfg_get(nrf_fem_control_cfg_t * p_cfg);
...
  *
- * This function will set up a timer to activate the pin 5 +/- 2.5 us before radio READY event is generated.
- * It will also set up a PPI to deactivate the pin on radio DISABLED event.
+ * */
+void nrf_fem_control_lna_ppi_disable(void);
+
+/**@brief Calculate target time for a timer, which activates the Power Amplifier pin.
+ * 
+ * @param[in] turnaround Information if TX mode is entered from RX mode (turnaround) or DISABLED (ramp up).
+ *
```
rlubos: Documentation for return value is missing.

src/fem/nrf_fem_control_api.h
```diff
@@ -125,54 +121,62 @@ void nrf_fem_control_cfg_get(nrf_fem_control_cfg_t * p_cfg);
...
+ * */
+uint32_t nrf_fem_control_pa_delay_get(bool turnaround);
+
+/**@brief Calculate target time for a timer, which activates the Low Noise Amplifier pin.
```
rlubos: Documentation for return value is missing.

src/fem/nrf_fem_control_common.c
```diff
@@ -150,75 +162,93 @@ void nrf_fem_control_activate(void)
...
 {
-    uint32_t target_time;
+    uint32_t target_time = 1;
```
rlubos: Why `1`? I saw some comments from previous iteration by @hubertmis, but what's the story behind that? Wouldn't `0` be more intuitive in that case?

hubertmis: I think `0` should be ok in this function. It should be verified on hardware smoke tests.

src/fem/nrf_fem_control_api.h
```diff
@@ -175,8 +182,8 @@ uint32_t nrf_fem_control_lna_delay_get(void);
 #define nrf_fem_control_lna_ppi_enable(...)
 #define nrf_fem_control_pa_ppi_disable(...)
 #define nrf_fem_control_lna_ppi_disable(...)
-#define nrf_fem_control_pa_delay_get(...)  1
-#define nrf_fem_control_lna_delay_get(...) 1
+#define nrf_fem_control_pa_delay_get(...)    0
+#define nrf_fem_control_lna_delay_get(...)   0
```
hubertmis: `0` for PA is OK, but for LNA it won't work. LNA needs `1` here.

src/nrf_drv_radio802154_fsm.c
```diff
@@ -780,7 +793,6 @@ static void tx_terminate(void)
     nrf_radio_shorts_set(SHORTS_IDLE);
     nrf_radio_task_trigger(NRF_RADIO_TASK_DISABLE);
 
-    nrf_fem_control_lna_ppi_disable();
```
hubertmis: TX state is tricky. It probably will use LNA for CCA procedure and PA for transmission. But you can leave it as it is. TX path is scope of another task.

src/fem/nrf_fem_control_common.c
```diff
@@ -150,75 +160,84 @@ void nrf_fem_control_activate(void)
...
     }
+
+    return target_time;
 }
 
-void nrf_fem_control_lna_set(bool shorts_used)
+uint32_t nrf_fem_control_lna_delay_get(void)
 {
-    uint32_t target_time;
+    uint32_t target_time = 0;
```
hubertmis: I think we still need `1` here.

src/nrf_drv_radio802154_fsm.c
```diff
@@ -1880,6 +1892,8 @@ static inline void irq_crcok_state_rx(void)
                                            (uint32_t)nrf_radio_task_address_get(
                                                    NRF_RADIO_TASK_TXEN));
             // TODO: Set FEM PPIs
```
hubertmis: not `TODO` anymore :)

src/nrf_drv_radio802154_fsm.c
```diff
@@ -792,6 +792,8 @@ static void tx_ack_terminate(void)
         nrf_radio_int_disable(ints_to_disable);
         nrf_radio_shorts_set(SHORTS_IDLE);
         nrf_radio_task_trigger(NRF_RADIO_TASK_DISABLE);
+
+        nrf_fem_control_pa_ppi_disable();
```
hubertmis: Move this line out of `if` block. PPI should be disabled regardless timeslot state.

PR #49, opened 2018-02-01
Continuous Carrier implementation and some cleanup

hubertmis:


src/nrf_drv_radio802154_fsm.c
```diff
@@ -366,11 +372,15 @@ static uint8_t * rx_buffer_get(void)
 static void cca_configuration_update(void)
```
rlubos: I think this function should return false in case configuration update failed do to missing timeslot. Currently, respective API function always returns true.

src/nrf_drv_radio802154_fsm.c
```diff
@@ -2073,7 +2176,10 @@ static void irq_phyend_state_tx_frame(void)
 
         nrf_egu_event_clear(NRF_DRV_RADIO802154_EGU_INSTANCE, EGU_EVENT);
 
-        nrf_ppi_channel_enable(PPI_CH1);
+        // Enable PPI disabled by DISABLED event
+        nrf_ppi_channel_enable(PPI_EGU_RAMP_UP);
+
```
rlubos: Is this empty line intended?

src/nrf_drv_radio802154_fsm.c
```diff
@@ -2958,27 +3074,37 @@ radio_state_t nrf_drv_radio802154_fsm_state_get(void)
...
+    if (m_state != RADIO_STATE_SLEEP)
     {
-        state_set(RADIO_STATE_SLEEP);
-        // TODO: Exit continuous mode, perhaps reset radio as well
+        result = current_operation_terminate(term_lvl);
+
+        if (result)
+        {
+            state_set(RADIO_STATE_SLEEP);
+
```
rlubos: IMHO  this empty line is not needed.

src/nrf_drv_radio802154_fsm.c
```diff
@@ -2958,27 +3074,37 @@ radio_state_t nrf_drv_radio802154_fsm_state_get(void)
...
     {
-        state_set(((m_state == RADIO_STATE_SLEEP) || (m_state == RADIO_STATE_WAITING_TIMESLOT)) ?
-                RADIO_STATE_WAITING_TIMESLOT : RADIO_STATE_WAITING_RX_FRAME);
+        result = current_operation_terminate(term_lvl);
 
-        receive_begin(true);
+        if (result)
+        {
+            state_set(RADIO_STATE_RX);
+
```
rlubos: Same as line 3086.

src/nrf_drv_radio802154_fsm.c
hubertmis: Not really. We should update CCA configuration in `nrf_raal_timeslot_started()` function. CCA configuration is not preserved between timeslots.

src/nrf_drv_radio802154_fsm.c
hubertmis: Yes. I've added a comment.

src/nrf_drv_radio802154_fsm.c
rlubos: Ok, so do you need to check if timeslot is granted then?

hubertmis: Yes. The driver shall not modify RADIO registers out of timeslot.

PR #51, opened 2018-02-05
Add ACK frame filtering

rlubos:


src/nrf_drv_radio802154_fsm.c
```diff
@@ -2278,16 +2282,23 @@ static void irq_end_state_rx_ack(void)
...
 
-    // TODO: Check MHR match
-
-    // TODO: There is a problem with ACK payload
-    transmitted_frame_notify(p_ack_buffer->psdu,                // psdu
-                             rssi_last_measurement_get(),       // rssi
-                             RX_FRAME_LQI(p_ack_buffer->psdu)); // lqi;
+    if (ack_match)
+    {
+        // TODO: There is a problem with ACK payload
```
hubertmis: Is there still any problem with ACK payload? I hope it is already fixed :)

src/nrf_drv_radio802154_fsm.c
```diff
@@ -2278,16 +2282,23 @@ static void irq_end_state_rx_ack(void)
     rx_buffer_t * p_ack_buffer = mp_current_rx_buffer;
     mp_current_rx_buffer->free = false;
```
hubertmis: I feel memory leak. This should be set only `if (ack_match)`.

src/nrf_drv_radio802154_fsm.c
rlubos: Well this `TODO` was quite generic, wasn't sure what it refers to so I've kept it. I'll remove it then.

PR #55, opened 2018-02-07
PSDU reception detector

rlubos:


src/nrf_drv_radio802154_fsm.c
```diff
@@ -1173,12 +1259,28 @@ static void receive_begin(bool disabled_was_triggered)
...
+                                           NRF_TIMER_TASK_CLEAR));
+#endif // NRF_DRV_RADIO802154_DISABLE_BCC_MATCHING
 
     nrf_ppi_channel_enable(PPI_EGU_RAMP_UP);
     nrf_ppi_channel_enable(PPI_EGU_TIMER_START);
     nrf_ppi_channel_enable(PPI_CRCERROR_CLEAR);
     nrf_ppi_channel_enable(PPI_CRCOK_DIS_PPI);
     nrf_ppi_channel_enable(PPI_DISABLED_EGU);
+#if NRF_DRV_RADIO802154_DISABLE_BCC_MATCHING
+    nrf_ppi_channel_enable(PPI_ADDRESS_COUNTER_COUNT);
```
hubertmis: These PPIs shall be enabled before `PPI_DISABLED_EGU`

src/nrf_drv_radio802154_fsm.c
```diff
@@ -1954,6 +2058,11 @@ static inline void irq_crcok_state_rx(void)
             nrf_timer_task_trigger(NRF_DRV_RADIO802154_TIMER_INSTANCE, NRF_TIMER_TASK_STOP);
             nrf_timer_task_trigger(NRF_DRV_RADIO802154_TIMER_INSTANCE, NRF_TIMER_TASK_CLEAR);
 
+#if NRF_DRV_RADIO802154_DISABLE_BCC_MATCHING
+            nrf_timer_task_trigger(NRF_DRV_RADIO802154_COUNTER_TIMER_INSTANCE, NRF_TIMER_TASK_STOP);
+            nrf_timer_task_trigger(NRF_DRV_RADIO802154_COUNTER_TIMER_INSTANCE, NRF_TIMER_TASK_CLEAR);
```
hubertmis: I think that PPIs for PSDU detection should be disabled in this handler. Is it omitted intentionally?

src/nrf_drv_radio802154_fsm.c
```diff
@@ -1151,13 +1224,26 @@ static void receive_begin(bool disabled_was_triggered)
...
+    nrf_ppi_channel_and_fork_endpoint_setup(PPI_EGU_TIMER_START,
+                                            (uint32_t)nrf_egu_event_address_get(
+                                                    NRF_DRV_RADIO802154_EGU_INSTANCE,
+                                                    EGU_EVENT),
+                                            (uint32_t)nrf_timer_task_address_get(
+                                                    NRF_DRV_RADIO802154_TIMER_INSTANCE,
+                                                    NRF_TIMER_TASK_START),
+                                            (uint32_t)nrf_timer_task_address_get(
+                                                    NRF_DRV_RADIO802154_COUNTER_TIMER_INSTANCE,
+                                                    NRF_TIMER_TASK_START));
```
hubertmis: It looks that `TIMER` is now working in two modes simulatenously: It is started in timer mode by this fork. It is also counting ADDRESS events and is cleared on CRCERROR event. I don't understand how is it supposed to work.

src/nrf_drv_radio802154_fsm.c
```diff
@@ -745,13 +784,23 @@ static void rx_terminate(void)
     nrf_ppi_channel_disable(PPI_CRCOK_DIS_PPI);
     nrf_ppi_channel_disable(PPI_EGU_RAMP_UP);
     nrf_ppi_channel_disable(PPI_EGU_TIMER_START);
+#if NRF_DRV_RADIO802154_DISABLE_BCC_MATCHING
+    nrf_ppi_channel_disable(PPI_ADDRESS_COUNTER_COUNT);
+    nrf_ppi_channel_disable(PPI_CRCERROR_COUNTER_CLEAR);
```
hubertmis: It looks that PPI fork configured in `receive_begin` function is not cleared here, nor in `CRCOK` event handler. It would lead to problems when this PPI is reused by other procedure.

src/nrf_drv_radio802154_fsm.c
jciupis: I don't get why `COUNTER_TIMER` would work in two modes simultaneously. The spec states that in both modes, the TIMER is started by triggering the START task.

hubertmis: Right. Thanks. I did not remember this part of the OPS.

rlubos: From infocenter:

> The TIMER can operate in two modes, Timer mode and Counter mode. In both modes, the TIMER is started by triggering the START task, and stopped by triggering the STOP task.

My understanding is, that in order to enable counting, you have start the timer. But I can be wrong of course.

src/nrf_drv_radio802154_fsm.c
rlubos: +1 for this change, that's what I was missing.

src/nrf_drv_radio802154_fsm.c
```diff
@@ -774,13 +823,23 @@ static void tx_ack_terminate(void)
     nrf_ppi_channel_disable(PPI_CRCOK_DIS_PPI);
     nrf_ppi_channel_disable(PPI_EGU_RAMP_UP);
     nrf_ppi_channel_disable(PPI_EGU_TIMER_START);
+#if NRF_DRV_RADIO802154_DISABLE_BCC_MATCHING
+    nrf_ppi_channel_disable(PPI_ADDRESS_COUNTER_COUNT);
```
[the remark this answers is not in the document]
rlubos: +1

src/nrf_drv_radio802154_fsm.c
rlubos: Could you elaborate? Why would order of enabling PPIs matter?

hubertmis: It may be possible that this function is interrupted between DISALBED_EGU is enabled and ADDRESS_COUNT is enabled. In this case we may skip COUNT task, because PPI is not configured before ADDRESS event is triggered. If all PPIs were enabled before DISABLED_EGU PPI, then frame reception cannot start before PPIs are enabled.

src/nrf_drv_radio802154_fsm.c
rlubos: Ok, it took me a while but I get it now what do you mean by 'clearing fork'. Not sure though if using forks in case PPIs are reused is a neat solution, but perhaps it's still better than using yet another PPI.

src/nrf_drv_radio802154_fsm.c
```diff
@@ -1904,6 +2008,14 @@ static inline void irq_crcok_state_rx(void)
 
             if (wait_for_phyend)
             {
+#if NRF_DRV_RADIO802154_DISABLE_BCC_MATCHING
+                nrf_ppi_fork_endpoint_setup(PPI_EGU_TIMER_START, 0);
```
hubertmis: This PPI should be disabled before PPI_TIMER_TX_ACK is set. 
If `wait_for_phyend` is false PPI should be enabled again. I think `receive_begin()` shall take care of that.

PR #56, opened 2018-02-07
FEM support for continuous carrier mode

jciupis:


src/fem/nrf_fem_control_common.c
```diff
@@ -44,10 +44,12 @@
...
 #include "hal/nrf_radio.h"
+#include "hal/nrf_timer.h"
 
-#define NRF_FEM_TIMER_INSTANCE         NRF_DRV_RADIO802154_TIMER_INSTANCE
-#define NRF_FEM_TIMER_LNA_CC_CHANNEL   0
-#define NRF_FEM_TIMER_PA_CC_CHANNEL    2
+#define NRF_FEM_TIMER_INSTANCE           NRF_DRV_RADIO802154_TIMER_INSTANCE
+#define NRF_FEM_TIMER_LNA_CC_CHANNEL     0
+#define NRF_FEM_TIMER_PA_CC_CHANNEL      2
+#define NRF_FEM_TIMER_PA_SHORT_STOP_MASK NRF_FEM_TIMER_PA_CC_CHANNEL
```
hubertmis: I'm not sure if it is good idea to define short mask as CC channel. They may have the same value in current chip revision. But if it changes it will lead to a bug. Hard-to-find bug

src/nrf_drv_radio802154_fsm.c
```diff
@@ -1450,16 +1453,30 @@ static void continuous_carrier_begin(bool disabled_was_triggered)
...
+    // Set FEM
+    nrf_fem_control_pa_ppi_enable();
+    nrf_fem_control_pa_timer_set();
+
     // Clr event EGU
     nrf_egu_event_clear(NRF_DRV_RADIO802154_EGU_INSTANCE, EGU_EVENT);
 
     // Set PPIs
+#if ENABLE_FEM
+    nrf_ppi_channel_and_fork_endpoint_setup(PPI_EGU_RAMP_UP,
```
hubertmis: It would be great if there were a function in FEM module that sets only fork of this PPI in case FEM is enabled. It would allow us to get rid of this `#if ENABLE_FEM` in fsm module.
Also, this fork must be cleared on `continuous_carrier_terminate()`. I wish there were a function in FEM that would clear previously set fork.

PR #59, opened 2018-02-12
FEM support for FSM TX path

jciupis:


src/nrf_drv_radio802154_fsm.c
```diff
@@ -2485,7 +2514,26 @@ static void irq_phyend_state_tx_frame(void)
...
+        // Set PPIs necessary in rx_ack state
+        nrf_fem_control_ppi_task_setup(NRF_FEM_CONTROL_LNA_PIN,
+                                       PPI_EGU_TIMER_START,
+                                       (uint32_t)nrf_egu_event_address_get(
+                                               NRF_DRV_RADIO802154_EGU_INSTANCE,
+                                               EGU_EVENT),
+                                       (uint32_t)nrf_timer_task_address_get(
+                                               NRF_DRV_RADIO802154_TIMER_INSTANCE,
+                                               NRF_TIMER_TASK_START));
+        nrf_fem_control_timer_set(NRF_FEM_CONTROL_LNA_PIN,
```
hubertmis: Is this timer reset in `rx_ack_terminate()`?

src/fem/nrf_fem_control_common.c
```diff
@@ -297,3 +297,41 @@ void nrf_fem_control_ppi_fork_clear(nrf_fem_control_pin_t pin, nrf_ppi_channel_t
...
+        {
+            pa_task_addr = (uint32_t)(&NRF_GPIOTE->TASKS_SET[m_nrf_fem_control_cfg.pa_gpiote_ch_id]);
+        }
+        else
+        {
+            pa_task_addr = (uint32_t)(&NRF_GPIOTE->TASKS_CLR[m_nrf_fem_control_cfg.pa_gpiote_ch_id]);
+        }
+    }
+    
+    nrf_ppi_channel_and_fork_endpoint_setup(ppi_channel, event_addr, lna_task_addr, pa_task_addr);
```
rlubos: In case PA or LNA is not enabled you will set rubbish here. You should initialize task addresses to 0. And if there is a chance to call this function when none is enabled, you should wrap ppi setup within `if` statement.

src/nrf_drv_radio802154_fsm.c
```diff
@@ -82,6 +82,7 @@
 #define PPI_EGU_RAMP_UP             PPI_CH1
 #define PPI_EGU_TIMER_START         PPI_CH2
 #define PPI_CRCERROR_CLEAR          PPI_CH3
+#define PPI_EGU_FEM_CCA             PPI_CH3
```
hubertmis: Why `EGU_FEM_CCA`. I think this PPI has nothing in commong with EGU. I would call it `PPI_CCAIDLE_FEM` - it connects CCAIDLE event with FEM

PR #63, opened 2018-02-13
Track in dirver FSM if RAAL is in continuous mode during falling asleep.

hubertmis:


src/nrf_drv_radio802154_fsm.c
```diff
@@ -1197,12 +1194,19 @@ static bool ppi_egu_worked(void)
...
+    nrf_drv_radio802154_priority_drop_timeslot_exit();
+}
+
+/** Begin Falling Asleep operation. */
+static void falling_asleep_begin(void)
 {
     if (!nrf_raal_timeslot_is_granted())
     {
-        sleep_start();
+        state_set(RADIO_STATE_SLEEP);
```
LuDuda: Dont you want put state_set function inside sleep_begin?

hubertmis: I was thinking about it, but other `_begin` functions do not update state. I'd like to keep it the same for all functions. Perhaps we can change it in future to include state update in `_begin` functions.

PR #66, opened 2018-02-14
Cleanup part 1

hubertmis:
Still I'm going to add errno module to clean error codes management. Also FSM module rename will follow in next PR.


src/nrf_drv_radio802154_fsm.c
```diff
@@ -217,21 +194,10 @@ static nrf_radio802154_flags_t m_flags;  ///< Flags used to store current driver
  *
  * @param[in]  state  Driver state to set.
  */
-static inline void state_set(radio_state_t state)
+static void state_set(radio_state_t state)
```
rlubos: What's wrong with `inline`?

src/nrf_drv_radio802154_fsm.c
```diff
@@ -762,70 +581,245 @@ static inline bool ed_iter_setup(uint32_t time_us)
  * @section FSM transition request sub-procedures
  **************************************************************************************************/
 
-/** Terminate Falling Asleep procedure. */
-static void falling_asleep_terminate(void)
+/** Wait necessary time to make sure PPIs and EGU marked event if it was triggered. */
```
rlubos: I don't get this comment, what does it mean?

src/nrf_drv_radio802154_fsm.c
rlubos: I would be consistent throghout the entire file then.

PR #68, opened 2018-02-15
Fix assertion in Ack Timeout when channel is busy.

hubertmis:


hubertmis:
I've renamed a variable to make code easier to understand.

PR #75, opened 2018-02-16
Refactoring of the driver core

hubertmis:


hubertmis:
Regression test

hubertmis:
Regression test

PR #78, opened 2018-02-20
Fix FEM run-time configuration

hubertmis:


src/fem/nrf_fem_control.c
```diff
@@ -248,48 +255,69 @@ uint32_t nrf_fem_control_delay_get(nrf_fem_control_pin_t pin)
...
+    if (pin_is_enabled(pin))
+    {
+        uint32_t target_time = nrf_fem_control_delay_get(pin);
+        nrf_timer_shorts_enable(NRF_FEM_TIMER_INSTANCE, short_mask);
+        nrf_timer_cc_write(NRF_FEM_TIMER_INSTANCE, timer_cc_channel, target_time);
+    }
 }
 
-void nrf_fem_control_timer_reset(nrf_timer_short_mask_t short_mask)
+void nrf_fem_control_timer_reset(nrf_fem_control_pin_t  pin, nrf_timer_short_mask_t short_mask)
```
KaSroka: Double space between nrf_fem_control_pin_t and pin

src/fem/nrf_fem_control_api.h
```diff
@@ -172,43 +175,49 @@ void nrf_fem_control_pin_clear(void);
...
 
 /**@brief Clear timer configuration after one of the Front End Module pins deactivation.
  * 
+ * @param[in] pin              The Front End Module controlled pin that was deactivated.
  * @param[in] short_mask       Mask of timer shortcuts to be disabled.
  *
- * */
-void nrf_fem_control_timer_reset(nrf_timer_short_mask_t short_mask);
+ */
+void nrf_fem_control_timer_reset(nrf_fem_control_pin_t  pin, nrf_timer_short_mask_t short_mask);
```
KaSroka: Another double space

KaSroka:
Isn't it strange to pass pin to _timer_reset function or similar?

hubertmis:
@KaSroka It is. That's why I removed it before. But setting / resetting timer is related to LNA or PA path and in case one of pins is disabled we shall not set / reset timer at some procedures.

KaSroka:
So shouldn't whole path for PA/LNA be if'ed on earlier stage and not just before touching hardware? Now with disabled pin we are calling some procedures which are eventually doing nothing. I prefer to do nothing as early as possible.

hubertmis:
Driver core is not aware of FEM configuration. It just calls FEM functions in appropriate places. FEM decides if PPI or TIMER should be configured according to FEM configuration.

KaSroka:
What if we add nrf_fem_control_pa_is_enabled and complementary function for lna and then move ifs to fsm?

hubertmis:
It adds more code to core. Moreover there are some function that uses both LNA and PA. It is easier to handle it in FEM module.

KaSroka:
Ok, I can live with that. So just please fix those two comments.

PR #80, opened 2018-03-01
Prevent notification of invalid frame on dst addr mismatch

hubertmis:


src/nrf_802154_core.c
```diff
@@ -1697,8 +1699,9 @@ static void irq_bcmatch_state_rx(void)
             rx_terminate();
             rx_init(true);
 
-            if ((mp_current_rx_buffer->psdu[FRAME_TYPE_OFFSET] & FRAME_TYPE_MASK) !=
-                FRAME_TYPE_ACK)
+            if ((filter_result != NRF_802154_RX_ERROR_INVALID_DEST_ADDR) &&
```
rlubos: OpenThread keeps track of how many frames were filtered due to destination address mismatch, so not reporting this would disable this funtionality.

hubertmis: I wasn't aware of this. Fixed.

PR #87, opened 2018-03-16
Correct RSSI, ED, LQI, CCA ED threshold values with temperature

hubertmis:
This PR adds new platform API that should be provided to allow the driver correctly calculate values based on RSSI measurements.


src/nrf_802154_rssi.c
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    if (temp <= 30)
+    {
+        return 0;
+    }
+
+    if (temp <= 50)
+    {
+        return - 1;
```
LuDuda: nit: remove extra spaces aftrer "-" char?

src/nrf_802154_rssi.c
```diff
@@ -0,0 +1,96 @@
...
+int8_t nrf_802154_rssi_sample_temp_corr_value_get(void)
+{
+    int8_t temp = nrf_802154_temperature_get();
+
+    if (temp <= -30)
+    {
+        return 3;
+    }
+
+    if (temp <= -10)
```
LuDuda: nit: why not else if?

src/nrf_802154_rssi.c
```diff
@@ -0,0 +1,96 @@
...
+}
+
+uint8_t nrf_802154_rssi_sample_corrected_get(uint8_t rssisample)
+{
+    return rssisample + nrf_802154_rssi_sample_temp_corr_value_get();
+}
+
+uint8_t nrf_802154_rssi_lqi_corrected_get(uint8_t lqi)
+{
+    return lqi - nrf_802154_rssi_sample_temp_corr_value_get();
```
LuDuda: Are we sure that we have same relation between LQI and temperature, as we have with RSSI?

I guess below ED and CCA_ED are correctly implemented.

src/nrf_802154_core.c
```diff
@@ -222,7 +223,11 @@ static void rx_flags_clear(void)
  */
 static int8_t rssi_last_measurement_get(void)
 {
-    return -((int8_t)nrf_radio_rssi_sample_get());
+    uint8_t rssisample = nrf_radio_rssi_sample_get();
```
LuDuda: I understand that register is called `RSSISAMPLE`, though, i would use rssi_sample for variable name everywhere possible.

Note: Your function already has this pattern: `nrf_radio_rssi_sample_get()`

src/nrf_802154_rssi.c
KaSroka: Or something like that:
```c
if ( temp <= -30)
{
    return 3;
}
else if (temp > 70)
{
    return -3;
}
else
{
    return (temp + 29)/20 - 2; //equivalent of return temp < 0 ? FLOOR((Temp - 30) / 20) : CEIL((temp - 30) / 20);
}
```

LuDuda: @KaSroka i would leave more if/else ifs statements personally, so that someone can easier understand implementation of errata.

hubertmis: I like if-else. I'll update it.

src/nrf_802154_rssi.c
hubertmis: Pretty sure. Check FTPAN-153 comments on Jira.

src/nrf_802154_core.c
hubertmis: @KaSroka, probably we should rethink `nrf_802154_rssi_last_get`. It is not tested right now and I'm afraid there are some bugs in this implementation.

src/nrf_802154_rssi.c
LuDuda: Yes i understand that LQI correction is needed, what i meant is to be sure that LQI increases/decreases with temperature in the same way as RSSI. So far we don't have indication that the formula is different, but i don't see a clear indication of oposite as well.

src/nrf_802154_rssi.c
```diff
@@ -0,0 +1,96 @@
...
+    {
+        result = -3;
+    }
+
+    return result;
+}
+
+uint8_t nrf_802154_rssi_sample_corrected_get(uint8_t rssisample)
+{
+    return rssisample + nrf_802154_rssi_sample_temp_corr_value_get();
```
LuDuda: nit: I would change rssisample to rssi_sample also here as well.

PR #88, opened 2018-03-20
Fix compilation problems with NDEBUG flag enabled

hubertmis:


nRF-IEEE-802.15.4-radio-driver.packsc
```diff
@@ -332,5 +332,24 @@
...
+        "_class": "nRF_drv_radio_802_15_4",
+        "_description": "This option enables NDEBUG compile option that removes asserts.",
+        "_defines": [
+            "NDEBUG"
+        ]
+    },
+
+    "lto": {
+        "_class": "nRF_drv_radio_802_15_4",
+        "_description": "This option enables Link Time Optimizations for driver applications build.",
```
rlubos: Not sure if `Link Time Optimizations` needs to be capitalized.

hubertmis: It is capitalized here: https://gcc.gnu.org/wiki/LinkTimeOptimization . I assumed it should be like this.

PR #91, opened 2018-03-29
Notify failure to mac_feature modules when entering critical section fails.

hubertmis:


src/nrf_802154_request_direct.c
```diff
@@ -84,14 +70,18 @@ void nrf_802154_request_init(void)
...
+    REQUEST_FUNCTION(nrf_802154_core_sleep, term_lvl)
 }
 
 bool nrf_802154_request_receive(nrf_802154_term_t              term_lvl,
                                 req_originator_t               req_orig,
                                 nrf_802154_notification_func_t notify_function)
 {
-    REQUEST_FUNCTION_3_ARGS(nrf_802154_core_receive, term_lvl, req_orig, notify_function)
+    REQUEST_FUNCTION_WITH_FAIL_INSTR(nrf_802154_core_receive,
+                                     notify_function(false); ,
```
KaSroka: I don't like this but I can't think of better idea right now

PR #95, opened 2018-04-03
Add function to calculate CCA threshold from given dBm.

hubertmis:


src/nrf_802154.c
```diff
@@ -150,6 +150,11 @@ int8_t nrf_802154_dbm_from_energy_level_calculate(uint8_t energy_level)
     return ED_MIN_DBM + (energy_level / ED_RESULT_FACTOR);
 }
 
+uint8_t nrf_802154_ccaedthres_from_dbm_calculate(int8_t dbm)
```
KaSroka: `ccaedthres` do we need all this information here and without underscores?

src/nrf_802154.h
```diff
@@ -211,6 +211,16 @@ void nrf_802154_short_address_set(const uint8_t *p_short_address);
  */
 int8_t nrf_802154_dbm_from_energy_level_calculate(uint8_t energy_level);
 
+/**
+ * @brief  Calculate CCA energy detection threshold value from given dBm.
```
b-gent: @brief  Calculate the CCA energy detection threshold value from a given dBm level.

@param[in]  dbm  Energy level in dBm used to calculate the CCAEDTHRES value.

@return  Energy level value corresponding to the given dBm level that should be written to the CCACTRL
register.

src/nrf_802154.c
KaSroka: Also I would omit `calculate` as `from_dbm` implies that it is somehow calculated from dBm.

hubertmis: CCAEDTHRES is name of field in the register. We calculate value that should be placed in this register.
FCS suggests to use verb to end function name.

PR #100, opened 2018-04-06
Fix deinitialization of the core module

hubertmis:


LuDuda:
I guess this is needed for switched multiprotocol solution. I would strongly recommend to update OT ASAP then.

hubertmis:
@LuDuda, You are right. I'm working on it.

PR #105, opened 2018-04-13
Use SHUTDOWN task for TIMERs to minimize power consumption in sleep

### LINKED ISSUE #103: High power consumption in sleep mode (27uA instead of 4uA)
Workaround for [anomaly 78](https://infocenter.nordicsemi.com/topic/com.nordic.infocenter.nrf52840.Rev1.errata/dita/errata/nRF52840/Rev1/latest/anomaly_840_78.html?cp=2_0_1_0_1_4)  is not implemented. Because of that TIMERs use too much power even if stopped when the driver is disabled.

hubertmis:
This PR fixes #103


hubertmis:
Regression test

PR #106, opened 2018-04-23
Refactor synchronization of timeslot exit between core and RAAL

hubertmis:


hubertmis:
Regression test

hubertmis:
Regression test

hubertmis:
Regression test

PR #107, opened 2018-04-25
Add initial version of Radio Scheduler

KaSroka:


src/nrf_802154_priority_drop_direct.c
```diff
@@ -46,7 +46,7 @@ void nrf_802154_priority_drop_init(void)
 
 void nrf_802154_priority_drop_timeslot_exit(void)
 {
-    nrf_raal_continuous_mode_exit();
```
hubertmis: I'm not sure about this change.
Perhaps core shall call `nrf_rsch_continuous_mode_exit()` directly, but RSCH should call `nrf_raal_continuous_mode_exit()` using priority drop module.

src/nrf_802154_swi.c
```diff
@@ -731,7 +731,7 @@ void SWI_IRQHandler(void)
 
     if (nrf_egu_event_check(SWI_EGU, TIMESLOT_EXIT_EVENT))
     {
-        nrf_raal_continuous_mode_exit();
```
hubertmis: This change is affected by comment in the priority drop module.

PR #110, opened 2018-04-26
Initial implementation of the delayed TX feature.

hubertmis:


src/mac_features/nrf_802154_delayed_trx.h
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+/**
+ * @brief Request transmission of a frame at given time.
+ *
+ * If CSMA-CA procedure is successful and frame is transmitted @sa nrf_802154_tx_started()
+ * function is called. If CSMA/CA procedure failed and frame cannot be transmitted due to busy
+ * channel @sa nrf_802154_transmit_failed() function is called.
+ *
+ * @note CSMA-CA does not timeout waiting for ACK automatically. Waiting for ACK shall be timed out
```
KaSroka: What about ACK timeout feature?

src/mac_features/nrf_802154_delayed_trx.c
```diff
@@ -0,0 +1,150 @@
...
+    return result;
+}
+
+void nrf_802154_rsch_delayed_timeslot_started(void)
+{
+    bool result;
+
+    assert(tx_is_in_progress());
+
+    // TODO: Set channel and reset it in the tx_started/tx_failed hook or if the request below failed.
```
KaSroka: Should this be implemented in this PR?

src/mac_features/nrf_802154_delayed_trx.c
```diff
@@ -0,0 +1,150 @@
...
+{
+    mp_tx_psdu = NULL;
+}
+
+/**
+ * Notify MAC layer that requested timeslot is not granted if tx request failed.
+ *
+ * @param[in]  result  Result of TX request.
+ */
+static void notify_tx_timeslot_denied(bool result)
```
KaSroka: This static function is not used

src/mac_features/nrf_802154_delayed_trx.c
hubertmis: I would add it later. It is not needed for alpha version.

src/mac_features/nrf_802154_delayed_trx.c
hubertmis: It is used :)

src/mac_features/nrf_802154_delayed_trx.h
hubertmis: Whole description here is invalid. Probably it is copy-paste. I'll update it.

src/mac_features/nrf_802154_delayed_trx.c
KaSroka: Oh, it's passed as a callback...

KaSroka: Oh, it's passed as a callback function...

PR #111, opened 2018-04-26
Move HF clock management from RAAL to RSCH

KaSroka:


src/nrf_802154_rsch.c
```diff
@@ -1,7 +1,129 @@
...
+{
+    __DMB();
+    m_mutex = 0;
+}
+
+static inline void prec_granted_set(rsch_prec_t prec)
+{
+    do
+    {
+        volatile rsch_prec_state_t old_state = __LDREXB(&m_prec_states[prec]);
```
hubertmis: Do we need to declare `old_state` as `volatile`?

src/nrf_802154_rsch.c
```diff
@@ -1,7 +1,129 @@
...
+
+    do
+    {
+        if (!mutex_trylock())
+        {
+            m_mutex_monitor++;
+            return;
+        }
+
+        temp_mon   = m_mutex_monitor;
```
hubertmis: There is a race if higher priority context calls this function between `mutex_trylock()` and `temp_mon = m_mutex_monitor` instructions.

src/nrf_802154_rsch.c
```diff
@@ -37,22 +177,33 @@ uint32_t nrf_802154_rsch_timeslot_us_left_get(void)
...
+    nrf_802154_log(EVENT_TRACE_EXIT, FUNCTION_RSCH_TIMESLOT_STARTED);
 }
 
-void nrf_raal_timeslot_started(void)
+void nrf_raal_timeslot_ended(void)
 {
-    nrf_802154_rsch_timeslot_started();
+    nrf_802154_log(EVENT_TRACE_ENTER, FUNCTION_RSCH_TIMESLOT_ENDED);
+
+    prec_requested_set(RSCH_PREC_RAAL);
```
hubertmis: This call should check if in the meantime the `cont_mode_exit` function changed state to `IDLE`. It should not set state to `REQUESTED` if it is `IDLE`.

src/nrf_802154_rsch.h
```diff
@@ -57,6 +57,22 @@ extern "C" {
  * granted access to the RADIO peripheral.
  */
 
+typedef enum
```
hubertmis: Are these enums in API of this module? Why are they declared in the header file?

src/nrf_802154_rsch.c
```diff
@@ -1,7 +1,165 @@
...
+            switch(curr_state)
+            {
+                case RSCH_PREC_GRANTED:
+                    nrf_802154_rsch_timeslot_started();
+                    break;
+                case RSCH_PREC_REVOKED:
+                    nrf_802154_rsch_timeslot_ended();
+                    break;
+                default:
+                    break;
```
hubertmis: Perhaps we could add `assert(false)` here?

src/nrf_802154_rsch.c
```diff
@@ -14,12 +172,30 @@ void nrf_802154_rsch_uninit(void)
...
 {
+    nrf_802154_log(EVENT_TRACE_ENTER, FUNCTION_RSCH_CONTINUOUS_EXIT);
+
+    prec_idle_set(RSCH_PREC_HFCLK);
+    nrf_802154_clock_hfclk_stop();
+
+    prec_idle_set(RSCH_PREC_RAAL);
     nrf_raal_continuous_mode_exit();
+
+    notify_core();
```
hubertmis: I would add a comment here that this function is just to set `m_last_state` to correct value. Perhaps `notify_core` could take `bool notify` argument and in case of `false` value set `m_last_state` value in the mutex section, but do not notify core module.

hubertmis: Or `bool in_continuous_mode` argument or any better name :)

KaSroka: Do we want to add this argument to this already complicated function? As we discussed earlier f2f it is ok to call timeslot_ended or timeslot_started after exiting continuous mode.

hubertmis: Calling `timeslot_ended` or `_started` out of continuous mode is safe, but it executes unnecessary code in the core module. We need to change this function a little in the next step when delayed TX is implemented in the RSCH module. We can postpone this modification until then.

PR #113, opened 2018-04-27
Handle critical sections in the core module.

hubertmis:


hubertmis:
Regression test

src/nrf_802154_critical_section.c
```diff
@@ -61,7 +61,7 @@ static volatile int8_t  m_nested_critical_section_allowed_priority;  ///< Indica
  */
 static void radio_critical_section_enter(void)
 {
-    if (nrf_802154_rsch_timeslot_is_granted())
+    if (nrf_raal_timeslot_is_granted())
```
KaSroka: Shouldn't we use RSCH here as we wanted to access RAAL only via RSCH?

src/nrf_802154_rsch.h
```diff
@@ -131,14 +131,6 @@ bool nrf_802154_rsch_timeslot_request(uint32_t length_us);
  */
 bool nrf_802154_rsch_delayed_timeslot_request(uint32_t t0, uint32_t dt, uint32_t length);
 
-/**
```
KaSroka: According to previous comment we should probably modify this function to return true only if RSCH preconditions are met and also modify it name?

KaSroka: Of course first part of this comment is valid for #111

src/nrf_802154_critical_section.c
hubertmis: The thing is that entering critical section should disable RADIO IRQ regardless HFCLK state and other preconditions. It should only check RAAL. Do you think that we should add a function like `nrf_802154_rsch_raal_timeslot_is_granted()`? Instead of calling RAAL directly from the critical section module?

KaSroka: I would make enum with preconditions (rsch_prec_t) public in #111 and add function like nrf_802154_rsch_prec_is_granted(rsch_prec_t prec); We may not need checking for other preconditions but it looks cleaner for me. As this PR goes in before #111 I would leave call to nrf_802154_rsch_timeslot_is_granted() (as all it does is calling RAAL counterpart) thus removing RAAL dependency and implement this in #111 properly.

PR #117, opened 2018-05-21
Delayed tx feature fixes after tests

hubertmis:


src/mac_features/nrf_802154_ack_timeout.c
```diff
@@ -61,46 +66,80 @@ static void notify_tx_error(bool result)
...
+    m_timer_idx = next_idx;
+
+    nrf_802154_timer_sched_add(p_timer, true);
+}
+
+static void timeout_timer_retry(nrf_802154_timer_t * p_prev_timer)
+{
+    uint32_t dt = p_prev_timer->dt + RETRY_DELAY;
+
+    assert(dt <= MAX_RETRY_DELAY);
```
LuDuda: I know asserts are removed in production code, though.. do you think it is a good idea to assert in this situation? Shouldn't we try until it succeed?

src/timer_scheduler/nrf_802154_timer_sched.c
```diff
@@ -396,19 +409,29 @@ bool nrf_802154_timer_sched_is_running(nrf_802154_timer_t * p_timer)
...
-        if (timer_shall_be_handled)
+        if (p_timer != NULL)
         {
-            handle_timer();
+            nrf_802154_timer_callback_t callback  = p_timer->callback;
+            void                      * p_context = p_timer->p_context;
+
+            (void)timer_remove(p_timer);
+
+            if ((callback != NULL))
```
LuDuda: nit: remove extra brackets

src/mac_features/nrf_802154_ack_timeout.c
hubertmis: I had a problem during tests with high-priority preemption that this timer was retrying to set up forever. Without this assert it was hard to find what is cause of infinity loop.

LuDuda: What about then to set timeout to some bigger value? My use case is that someone really tries to do sth long in interrupt handler e.g.  initilize SoftDevice (300+ms). I guess not all software is built with NDEBUG flag.

PR #118, opened 2018-05-23
Add timer to critical sections

hubertmis:


src/nrf_802154_critical_section.c
```diff
@@ -125,14 +217,63 @@ static bool critical_section_enter(bool forced)
...
 
-    nrf_802154_log(EVENT_TRACE_EXIT, FUNCTION_CRIT_SECT_ENTER);
     return result;
 }
 
+static void critical_section_exit(void)
+{
+    uint8_t     cnt;
+    static bool exiting_crit_sect;
+    rsch_evt_t  rsch_evt= rsch_pending_evt_clear();
```
pszkotak: nit: space missing

src/nrf_802154_critical_section.c
```diff
@@ -125,14 +217,63 @@ static bool critical_section_enter(bool forced)
...
     while (__STREXB(cnt + 1, &m_nested_critical_section_counter));
 
-    nrf_802154_log(EVENT_TRACE_EXIT, FUNCTION_CRIT_SECT_ENTER);
     return result;
 }
 
+static void critical_section_exit(void)
+{
+    uint8_t     cnt;
+    static bool exiting_crit_sect;
```
LuDuda: nit: Maybe we can move it to the top of the file?

src/nrf_802154_swi.h
```diff
@@ -164,6 +164,9 @@ void nrf_802154_swi_receive(nrf_802154_term_t              term_lvl,
  * @param[in]   req_orig         Module that originates this request.
  * @param[in]   p_data           Pointer to PSDU of the frame to transmit.
  * @param[in]   cca              If the driver should perform CCA procedure before transmission.
+ * @param[in]  immediate         If true, the driver schedules transmission immediately or never;
```
LuDuda: nit: alignment

src/nrf_802154_critical_section.c
hubertmis: I don't think so. It is local static variable. It should be encapsulated in a function. There is no need to extend its scope.

PR #122, opened 2018-05-29
Rename Timer to LP Timer

KaSroka:


src/mac_features/nrf_802154_ack_timeout.c
```diff
@@ -42,17 +42,17 @@
 
 #include "nrf_802154_notification.h"
 #include "nrf_802154_request.h"
-#include "timer_scheduler/nrf_802154_timer_sched.h"
+#include "timer_scheduler/nrf_802154_lp_timer_sched.h"
```
hubertmis: Is it intentional to rename this module?

src/mac_features/nrf_802154_ack_timeout.c
```diff
@@ -42,17 +42,17 @@
...
 #define MAX_RETRY_DELAY 1000000  ///< Maximal allowed delay of procedure retry.
 
 static void timeout_timer_retry(void);
 
-static uint32_t           m_timeout = NRF_802154_ACK_TIMEOUT_DEFAULT_TIMEOUT;  ///< ACK timeout in us.
-static nrf_802154_timer_t m_timer;                                             ///< Timer used to notify when we are waiting too long for ACK.
-static volatile bool      m_procedure_is_active;
-static const uint8_t    * mp_frame;
+static uint32_t              m_timeout = NRF_802154_ACK_TIMEOUT_DEFAULT_TIMEOUT;  ///< ACK timeout in us.
+static nrf_802154_lp_timer_t m_timer;                                             ///< Timer used to notify when we are waiting too long for ACK.
```
hubertmis: This type is defined in the timer scheduler module. I think it should be independent from lp/hp timers.

src/nrf_802154_config.h
```diff
@@ -124,13 +124,13 @@ extern "C" {
 #endif
 
 /**
- * @def NRF_802154_TIMER_INSTANCE
+ * @def NRF_802154_LP_TIMER_INSTANCE
```
hubertmis: This is independent from LP_TIMER. Should not be renamed.
This timer is used solely by the core module.

src/platform/timer/nrf_802154_lp_timer.h
```diff
@@ -58,54 +58,54 @@ extern "C" {
  * * CSL
  * * Auto polling by rx-off-when-idle devices
  *
- * @note Most of Timer Abstraction Layer API should not be called directly by 802.15.4 driver
+ * @note Most of Low Power Timer Abstraction Layer API should not be called directly by 802.15.4 driver
  *       modules. This API is used by the Timer Scheduler module included in the driver and other
```
hubertmis: Perhaps we can add a note that API of this module is used by the timer coordinator module as well (the new module we are going to introduce soon)

KaSroka: So we want to add this note now or when we introduce this module?

hubertmis: I would add this note now. I'm not sure if we will remember about it during implementation of the time coord :) . But we can leave it if you would remember.

PR #123, opened 2018-05-29
API of the timer coordinator module

hubertmis:


src/nrf_802154_timer_coord.h
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * @param[in]  event_addr  Address of given event.
+ */
+void nrf_802154_timer_coord_timestamp_prepare(uint32_t event_addr);
+
+/**
+ * @brief Get timestamp of the recently prepared event.
+ *
+ * If recently prepared event occurred a few times since preparation, this function returns
+ * timestamp of the first occurrence.
```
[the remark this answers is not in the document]
hubertmis: I can imagine that we pass END event here in radio RX state. END event would be triggered when frame is received and when ACK is transmitted. We want to get timestamp of first occurrence.

src/nrf_802154_timer_coord.h
```diff
@@ -0,0 +1,114 @@
...
+void nrf_802154_timer_coord_init(void);
+
+/**
+ * @brief Uninitialize the Timer Coordinator module.
+ *
+ */
+void nrf_802154_timer_coord_uninit(void);
+
+/**
+ * @brief Start timer coordinator.
```
LuDuda: nit: align to above "the Timer Coordinator"

src/nrf_802154_timer_coord.h
```diff
@@ -0,0 +1,114 @@
...
+ * 
+ * This function starts the HP timer and synchronizes it to the LP timer.
```
LuDuda: nit: synchronize with?

src/nrf_802154_timer_coord.h
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+ * @brief Stop timer coordinator.
+ *
+ * This function stops the HP timer.
+ */
+void nrf_802154_timer_coord_stop(void);
+
+/**
+ * @brief Prepare getting precise timestamp of given event.
```
LuDuda: Is it address of hardware event e.g. EVENT_END? I guess so, then i would specify this in the dox.

src/nrf_802154_timer_coord.h
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+ *@}
+ **/
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* NRF_802154_RSCH_H_ */
```
LuDuda: NRF_802154_TIMER_COORD_H_

hubertmis:
@LuDuda I don't want to name this module `timestamp sth`, because it is intended to do much more than just timestamps. It coordinates HP timer with LP timer (that's where current name proposal came from :) ). The first feature of HP timer we are implementing is timestamp indeed. But hopefully we will implement more features and renaming would be a waste.

PR #125, opened 2018-05-30
Implementation of the HP Timer module

hubertmis:


src/platform/hp_timer/nrf_802154_hp_timer.c
```diff
@@ -0,0 +1,137 @@
...
+
+void nrf_802154_hp_timer_sync_prepare(void)
+{
+    uint32_t curr_time = timer_time_get();
+    uint32_t past_time = curr_time - 1;
+
+    m_unexpected_sync = past_time;
+}
+
+bool nrf_802154_hp_timer_sync_time_get(uint32_t * p_timestamp)
```
KaSroka: Assert if `p_timestamp` is NULL

PR #130, opened 2018-06-15
Added PIB coordinator flag and condition for filtering dst_mode_none frames on non-coordinator devices

AlexeyMatveevDSR:


src/nrf_802154.c
```diff
@@ -293,7 +293,6 @@ bool nrf_802154_transmit(const uint8_t * p_data, uint8_t length, bool cca)
                                          REQ_ORIG_HIGHER_LAYER,
                                          m_tx_buffer,
                                          cca,
-                                         false,
```
hubertmis: It seems that removing of this line is a regression. Could you restore this line in the PR?

src/mac_features/nrf_802154_filter.c
```diff
@@ -179,7 +179,7 @@ static bool dst_addressing_end_offset_get_2006(const uint8_t * p_psdu,
             }
             else
             {
-                result = true;
```
hubertmis: With this change all frames with None destination address are further filtered in `dst_pan_id_check()`, `dst_short_addr_check()` and `dst_extended_addr_check()`. But there are no new conditions in these functions that would accept frames with None address when the node is PAN Coordinator. It seems that with this patch all frames destined to a PAN Coordinator will be dropped regardless state of the node. Is it intentional limitation for now?

src/nrf_802154.c
AlexeyMatveevDSR: Done. Restored in the feature branch.

src/mac_features/nrf_802154_filter.c
AlexeyMatveevDSR: Agree. Added filter conditions in dst_short_addr_check and dst_extended_addr_check. PanID filter should not be modified.

AlexeyMatveevDSR: Tested by sending None destination address frame to ZC (accepted) and to ZR (rejected w/o ACK).

hubertmis:
@AlexeyMatveevDSR, could you rebase this branch to the master? Also it would be great if you squashed commits from this PR to single one.

hubertmis:
Regression test

hubertmis:
@AlexeyMatveevDSR, could you rebase to the master branch again? Sorry for the inconvenience, but we've found a problem in our test system and related bug in the stack itself. It caused fail of CI tests of this PR.

hubertmis:
Or perhaps I can rebase myself with Github GUI

PR #137, opened 2018-07-02
Refactor RSCH to support priority levels internally.

hubertmis:


src/rsch/nrf_802154_rsch.c
```diff
@@ -81,96 +80,89 @@ static bool any_prec_should_be_requested_for_delayed_timeslot(void)
...
-        }
-    } while (__STREXB((uint8_t)RSCH_PREC_STATE_APPROVED, (uint8_t*)&m_prec_states[prec]));
+    return result;
 }
 
-/** @brief Set RSCH_PREC_STATE_REQUESTED state on given precondition @p prec only if
- *         its current state is RSCH_PREC_STATE_APPROVED.
+/** @brief Set approved priority level @p prio on given precondition @p prec.
  *
+ * This function sets only @ref RSCH_PRIO_IDLE priority level if requested priority level is
```
KaSroka: Description may be incorrectly interpreted - we can set RSCH_PRIO_IDLE only when RSCH_PRIO_IDLE is requested. Changing order in sentence may prevent this. `When requested priority level equals to RSCH_PRIO_IDLE this function will approve only RSCH_PRIO_IDLE priority level and drop other approved levels silently.`

src/rsch/nrf_802154_rsch.c
```diff
@@ -180,79 +172,68 @@ static inline void all_prec_request(void)
...
  *
- * @retval true   All preconditions are requested or met.
- * @retval false  At least one precondition is idle.
+ * @param[in]  prio  Minimal priority level requested from preconditions.
+ *
+ * @retval true   All preconditions are requested or met at given level.
+ * @retval false  At least one precondition is requested at lower level than required.
  */
-static inline bool all_prec_are_requested(void)
+static inline bool all_prec_are_requested(rsch_prio_t prio)
```
KaSroka: I think that this function name is obsolete.

PR #140, opened 2018-07-09
Prevent reentrancy of SWI notifications

hubertmis:


src/nrf_802154_swi.c
```diff
@@ -322,6 +322,41 @@ static bool ntf_queue_is_empty(void)
...
+ * Enter notify block.
+ *
+ * This is a helper function used in all notification functions to atomically
+ * find an empty slot in the notification queue and allow atomic slot update.
+ *
+ * @return Pointer to an empty slot in the notification queue.
+ */
+static nrf_802154_ntf_data_t * ntf_enter(void)
+{
+    __disable_irq();
```
KaSroka: Can't we avoid global IRQ disable?

hubertmis: Probably we could, but it would be harder to implement and maintain. This module is used only with SD RAAL and because of this global IRQ disable is not a big deal.

PR #143, opened 2018-07-16
Implementing nrf_802154_sleep_if_idle() function.

IgorPivovarovDSR:
According to the polling mechanism stated in MAC specification, ED sends Data Request to a router, then  router replies with ACK. If pending bit is set ED start waiting for the packet for a certain constant period of time. Then ED is able to go to sleep.
However, if a timeout occurs during the reception of the packet, the driver will notify us by calling nrf_802154_receive_failed with NRF_DRV_RADIO802154_TX_ERROR_ABORTED code in that case. 

-----

The nrf_802154_sleep_if_idle() function should forbid radio to change state if PSDU is being recived.

### DIFF
--- a/src/nrf_802154.h
+++ b/src/nrf_802154.h
@@ -263,6 +263,19 @@ nrf_802154_state_t nrf_802154_state_get(void);
  */
 bool nrf_802154_sleep(void);
 
+/**
+ * @brief Change radio state to sleep if radio is idle.
+ *
+ * Sleep state is the lowest power state. In this state, the radio cannot transmit or receive
+ * frames. It is the only state in which the driver releases the high-frequency clock and does not
+ * request timeslots from a radio arbiter.
+ *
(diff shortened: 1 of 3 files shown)


src/nrf_802154.h
```diff
@@ -263,6 +263,8 @@ nrf_802154_state_t nrf_802154_state_get(void);
  */
 bool nrf_802154_sleep(void);
 
+nrf_802154_sleep_error_t nrf_802154_sleep_if_idle(void);
```
hubertmis: nit: please copy doxygen from the `nrf_80215_sleep()` and update `@retval` values.

hubertmis:
Regression test

hubertmis:
Regression test

hubertmis:
Regression test

hubertmis:
Regression test

PR #145, opened 2018-07-27
Fix margin calculation

KaSroka:


src/raal/softdevice/nrf_raal_softdevice.c
```diff
@@ -185,13 +184,38 @@ static inline bool timer_is_set_to_margin(void)
...
+}
+
+/**@brief Eliminate timers jitters. */
+static uint32_t timer_jitter_adjust(void)
+{
+    // Adjust TIMER0 and RTC0 clocks drifts.
+    uint32_t timer_ticks = timer_time_get();
+    uint32_t margin_cc;
+
+    margin_cc = timer_ticks + m_config.timeslot_length - m_config.timeslot_safe_margin;
```
LuDuda: I think we can't use m_config.timeslot_length directly here, because of one potential issue.

If we get NRF_EVT_RADIO_BLOCKED or NRF_EVT_RADIO_CANCELED we try to request smaller timeslot, e.g. with 3200us. And if right after on first extension request we get NRF_RADIO_CALLBACK_SIGNAL_TYPE_EXTEND_FAILED event then we incorrectly set margin to the moment in the future (after real timeslot end).

hubertmis: Also if our timeslot ends just after max timeslot length it may results in incorrect margin setting. It should be improved.

src/raal/softdevice/nrf_raal_softdevice.c
```diff
@@ -185,13 +175,27 @@ static inline bool timer_is_set_to_margin(void)
     return m_timer_action == TIMER_ACTION_MARGIN;
 }
 
+/**@brief Eliminate timers jitters. */
+static uint32_t timer_jitter_adjust(void)
```
LuDuda: Oh, actually i was thinking of renaming this function to timer_get_cc_margin instead of one above.

src/raal/softdevice/nrf_raal_softdevice.h
```diff
@@ -43,13 +43,35 @@
 extern "C" {
 #endif
 
+/**@brief RTC clock frequency. */
+#define RTC_FREQUENCY             32768ULL
```
LuDuda: nit: Those four next defines names are very generic, and they are defined in the header. Ideally i would move them all to .c file. Right now there is a chance of redefinitions while someone includes nrf_raal_softdevice.h in its project.

src/raal/softdevice/nrf_raal_softdevice.c
```diff
@@ -185,13 +175,27 @@ static inline bool timer_is_set_to_margin(void)
...
+
+    margin_cc = timer_ticks + m_last_timeslot_length - m_config.timeslot_safe_margin;
+    margin_cc -= TIMER_TO_SIGNAL_JITTER_US;
+
+    return margin_cc;
+}
+
 /**@brief Set timer action to the timeslot margin. */
-static inline void timer_to_margin_set(void)
+static inline void timer_get_cc_margin(void)
```
hubertmis: Name of this function is misleading. I expect this function to return cc_margin value, but its type is `void ()`

src/raal/softdevice/nrf_raal_softdevice.c
```diff
@@ -483,8 +430,8 @@ static nrf_radio_signal_callback_return_param_t *signal_handler(uint8_t signal_t
 
         assert(m_timeslot_state == TIMESLOT_STATE_REQUESTED);
 
-        m_start_rtc_ticks = NRF_RTC0->COUNTER;
         m_timeslot_state  = TIMESLOT_STATE_GRANTED;
```
hubertmis: nit: alignment

src/raal/softdevice/nrf_raal_softdevice.h
```diff
@@ -59,7 +81,6 @@ typedef struct
     uint32_t timeslot_max_length;    /**< Maximum single timeslot length created by extension processing in microseconds. */
     uint16_t timeslot_alloc_iters;   /**< Maximum number of iteration of dividing timeslot_length by factor of 2 performed by arbiter. */
     uint16_t timeslot_safe_margin;   /**< Safe margin before timeslot is finished and nrf_raal_timeslot_ended should be called in microseconds. */
-    uint16_t lf_clk_accuracy_ppm;    /**< Clock accuracy in ppm unit. */
```
hubertmis: Like discussed F2F, try to not modify external API. Just mark as deprecated.

src/raal/softdevice/nrf_raal_softdevice.h
```diff
@@ -46,20 +46,39 @@ extern "C" {
...
-    uint32_t timeslot_timeout;       /**< Longest acceptable delay until the start of the requested timeslot in microseconds. */
-    uint32_t timeslot_max_length;    /**< Maximum single timeslot length created by extension processing in microseconds. */
-    uint16_t timeslot_alloc_iters;   /**< Maximum number of iteration of dividing timeslot_length by factor of 2 performed by arbiter. */
-    uint16_t timeslot_safe_margin;   /**< Safe margin before timeslot is finished and nrf_raal_timeslot_ended should be called in microseconds. */
-    uint16_t lf_clk_accuracy_ppm;    /**< Clock accuracy in ppm unit. */
+    /**
+     * @brief Timeslot length requested by the module in microseconds.
+     */
+    uint32_t timeslot_length;
+    /**
```
LuDuda: nit: make a new line above?

src/raal/softdevice/nrf_raal_softdevice.c
```diff
@@ -185,13 +202,27 @@ static inline bool timer_is_set_to_margin(void)
     return m_timer_action == TIMER_ACTION_MARGIN;
 }
 
+/**@brief Eliminate timers jitters. */
```
LuDuda: nit: Change the comment.

src/raal/softdevice/nrf_raal_softdevice.h
```diff
@@ -39,27 +39,71 @@
...
+
+    /**
+     * @brief Safe margin before timeslot is finished and nrf_raal_timeslot_ended should be called in microseconds.
+     * @ref nrf_raal_softdevice_safe_margin_calc can be used to calculate proper value based on clock accuracy.
+     * This value can also be selected experimentally.
+     */
+    uint16_t timeslot_safe_margin; 
+
+    /**
+     * @brief @deprecated Clock accuracy in ppm unit.
```
hubertmis: Could you add a note that value of this field is not used anymore and ppm is included in the `timeslot_safe_margin` now?

PR #150, opened 2018-08-09
Optimize timing calculations

hubertmis:


src/nrf_802154_utils.h
```diff
@@ -50,11 +53,83 @@
...
+
+       (2^56)/15625) is equal to 0x431BDE82D7B, and is written in hexadecimal representation
+       to show the bit width of the number. Shifting 56 bits is determined by the worst
+       case value of t1. The constant is selected by using the same methodology as in the
+       first step of the calculation above.
+
+       The possible loss of precision in the calculation above can make t1 93*15625 lower
+       than it should have been here. The worst case found is that t1 can be 1453125, and
+       therefore there is no overflow in the calculation
+       1453125 * 0x431BDE82D7B = 0x5cfffffffff76627 (i.e. it is less than 0xFFFFFFFFFFFFFFFF).
```
jciupis: super nit: I would unify use of small and big letters in hex notation.

src/platform/lp_timer/nrf_802154_lp_timer_nodrv.c
```diff
@@ -56,20 +57,14 @@
 #define RTC_SYNC_COMPARE_EVENT          NRF_RTC_EVENT_COMPARE_1
 #define RTC_SYNC_COMPARE_EVENT_MASK     RTC_EVTEN_COMPARE1_Msk
 
-#define RTC_FREQUENCY                   32768ULL
-
-#define US_PER_S                        1000000ULL
-#define US_PER_TICK                     CEIL_DIV(US_PER_S, RTC_FREQUENCY)
-#define US_PER_OVERFLOW                 (512UL * US_PER_S)                  ///< Time that has passed between overflow events. On full RTC speed, it occurs every 512 s.
+#define US_PER_TICK                     NRF_802154_DIVIDE_AND_CEIL(NRF_802154_US_PER_S,            \
```
KaSroka: What about `NRF_802154_RTC_TICKS_TO_US(1)?`

KaSroka: And we can move it to utils.

hubertmis: I agree, but:
* macro is more prone to compile-time optimization than `NRF_802154_RTC_TICKS_TO_US` function;
* this value is used only in this file, we do not gain a lot by moving it to the utils file.

I would refactor it when we can gain something by this operation.

hubertmis:
Regression test

PR #153, opened 2018-08-14
Fix direct Front-End module pins reset

jciupis:
According to nRF52840 specification:

> As long as a SET[n], CLR[n] and OUT[n] task or an IN[n] event is configured to control a pin n, the pin's output value will only be updated by the GPIOTE module. The pin's output value as specified in the GPIO will therefore be ignored as long as the pin is controlled by GPIOTE. Attempting to write a pin as a normal GPIO pin will have no effect. 

This PR fixes clearing Front-End module pins, which is done incorrectly according to the chip spec.


src/fem/nrf_fem_control.c
```diff
@@ -257,14 +257,14 @@ void nrf_fem_control_pin_clear(void)
 {
     if (pin_is_enabled(NRF_FEM_CONTROL_PA_PIN))
     {
-        nrf_gpio_pin_write(m_nrf_fem_control_cfg.pa_cfg.gpio_pin,
-                           !m_nrf_fem_control_cfg.pa_cfg.active_high);
+        nrf_gpiote_task_force(m_nrf_fem_control_cfg.pa_gpiote_ch_id,
+                              (nrf_gpiote_outinit_t)!m_nrf_fem_control_cfg.pa_cfg.active_high);
```
KaSroka: This cast looks weird, do we really should cast from `bool` to `*init_t`?

hubertmis: I think that proper way would be to create static function or macro: `#define gpiote_outinit_get(bool active_high) (active_high) ? GPIOTE_CONFIG_OUTINIT_Low : GPIOTE_CONFIG_OUTINIT_High` and use it here. But I'm OK to leave it as it is

jciupis: It seems to be a bit of an overkill compared to the cast :)

hubertmis: Unless order in enum changes :)

KaSroka: Macro looks cleaner for me but it's an overkill probably.

PR #159, opened 2018-10-02
Return success for core transition request from rx state to tx state

hubertmis:
This PR also changes transition from any state to sleep state to
return success even if there is not enough time to perfrom transition
due to ending timeslot.


hubertmis:
Regression test

src/nrf_802154_core.c
```diff
@@ -2705,18 +2726,25 @@ bool nrf_802154_core_receive(nrf_802154_term_t              term_lvl,
...
+    bool result = nrf_802154_critical_section_enter();
 
     if (result)
     {
         if ((m_state != RADIO_STATE_RX) && (m_state != RADIO_STATE_TX_ACK))
         {
-            result = current_operation_terminate(term_lvl, req_orig, notify_abort);
+            if (critical_section_can_be_processed_now())
+            {
+                result = current_operation_terminate(term_lvl, req_orig, notify_abort);
```
LuDuda: When this can fail?

hubertmis: Termination of current operation? Most likely when termination level is too low.

hubertmis:
Regression test

PR #171, opened 2018-11-15
[style] make code pretty

LuDuda:
This PR fixes some style issues.

Generally i've experimented with clang-format before my absence. And i found it unusable (it's C++ oriented). IMHO adding it to CI will cause too many unwanted behavior and will be time consuming. 

There are multiple problems, e.g. with defines alignment (not supported), with variables alignment (strange behaviors), column width, and many many more.. And so I had to manually added certain modifications in the files.

For C beautifier i would consider other tool in the future.


hubertmis:
I encourage you to create valid script in proven C style tool instead of fixing problems manually. I think it would be waste of time :)

hubertmis:
I think I've seen an uncristify script in the SDK repository. It might be good to check.

LuDuda:
Well, unfortunately i have no time for that now, so if some improvements is worse than nothing i am closing this PR. Wanted to push it out since code is alive.

LuDuda:
Seems that uncrustify is much better, i see only minor issues. I will issue a new PR soon. Does it make sense to you to enable travis for this job?

hubertmis:
I think travis checking coding standard would be great!

PR #176, opened 2018-11-30
Improve handling of frames of incorrect length in promiscuous mode

pszkotak:


src/nrf_802154_types.h
```diff
@@ -76,13 +76,14 @@ typedef uint8_t nrf_802154_rx_error_t;
...
 #define NRF_802154_RX_ERROR_NONE              0x00 // !< There is no receive error.
 #define NRF_802154_RX_ERROR_INVALID_FRAME     0x01 // !< Received a malformed frame.
-#define NRF_802154_RX_ERROR_INVALID_FCS       0x02 // !< Received a frame with invalid checksum.
-#define NRF_802154_RX_ERROR_INVALID_DEST_ADDR 0x03 // !< Received a frame with mismatched destination address.
-#define NRF_802154_RX_ERROR_RUNTIME           0x04 // !< A runtime error occurred (for example, CPU was held for too long).
-#define NRF_802154_RX_ERROR_TIMESLOT_ENDED    0x05 // !< Radio timeslot ended during frame reception.
-#define NRF_802154_RX_ERROR_ABORTED           0x06 // !< Procedure was aborted by another driver operation with FORCE priority.
-#define NRF_802154_RX_ERROR_TIMESLOT_DENIED   0x07 // !< Delayed reception request was rejected due to denied timeslot request.
-#define NRF_802154_RX_ERROR_TIMEOUT           0x08 // !< Frame not received during delayed reception time slot.
+#define NRF_802154_RX_ERROR_INVALID_LENGTH    0x02 // !< Received a frame with invalid length.
```
hubertmis: Don't change enumeration already present in the public API to keep API compatibility. Add new indication at the end of the list.

hubertmis: Now this error has the same value as `INVALID_FCS`. I think it should be `0x09`

pszkotak: Sure, fixed ;).

PR #146, opened 2018-07-30
Extract Ack-related logic to a separate module

jciupis:

PR #179, opened 2018-12-07
Enhanced Acks generator

### LINKED ISSUE #177: Check destiny PAN Id function not working
I'm testing the driver to integrate with one of our products and the function static `bool dst_pan_id_check(const uint8_t * p_psdu)` is returning false where the input data is correct. 
The reason is that  `PAN_ID_OFFSET` is set to 4 and then when reading the data `&p_psdu[PAN_ID_OFFSET]` it reading from the wrong position. 

The comment says `///< Offset of Pan Id in Data frame (+1 for frame length byte).` but then when accessing the buffer it should decremented `PAN_ID_OFFSET` by 1. 

I've defining `PAN_ID_OFFSET` to 3 and now I'm receiving packets, but I'm not doing a pull request because  I'm not sure if this is going to impact somewhere else.

hubertmis:
This PR provides enhanced Ack support, better performance of the Precise Ack Timeout module and improved support of other features present in the 802.15.4-2015 specification.

This PR should fix #177.
I based this patch on #146.


src/nrf_802154_core_hooks.c
```diff
@@ -104,6 +105,15 @@ static const tx_started_hook m_tx_started_hooks[] =
     NULL,
 };
 
+static const rx_ack_started_hook m_rx_ack_started_hooks[] =
```
KaSroka: Future improvement possible: section variable

src/mac_features/ack_generator/nrf_802154_enh_ack_generator.c
```diff
@@ -174,78 +153,90 @@ static void frame_control_set(const uint8_t * p_frame, const uint8_t * p_ie_data
     fcf_frame_version_set();
     fcf_src_addressing_mode_set(p_frame);
 
-    m_ack_psdu[0] += FCF_SIZE;
-    m_ack_psdu[0] += FCS_SIZE;
+    parse_results = nrf_802154_frame_parser_mhr_parse(m_ack_psdu, p_ack_offsets);
+    assert(parse_results);
+    (void)parse_results;
+
+    m_ack_psdu[0] = p_ack_offsets->addressing_end_offset - 1 + FCS_SIZE;
```
aprochwicz: Just for consistency, maybe use PHR_OFFSET instead of magic number 0?

src/mac_features/ack_generator/nrf_802154_enh_ack_generator.c
```diff
@@ -133,37 +116,33 @@ static void fcf_dst_addressing_mode_set(const uint8_t * p_frame)
...
     }
-    else
+    else if (nrf_802154_frame_parser_src_addr_is_short(p_frame))
     {
         m_ack_psdu[DEST_ADDR_TYPE_OFFSET] |= DEST_ADDR_TYPE_SHORT;
-        m_ack_psdu[0] += SHORT_ADDRESS_SIZE;
+    }
+    else
+    {
+        m_ack_psdu[DEST_ADDR_TYPE_OFFSET] |= DEST_ADDR_TYPE_NONE;
```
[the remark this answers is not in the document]
hubertmis: The compiler will remove it anyway during optimizations. The reason I kept it is FCS 8.2.4:
> Any if statement with an else if clause shall end with an else clause.
I could create an empty `else` clause with appropriate comment. However I think current version is cleaner.

PR #188, opened 2019-01-16
Add a workaround for a CPU locks while BLE is scanning

tomchy:
Simple workaround proposed by @hubertmis for a bug reported as DRGN-10402.

This allows to avoid CPU locks if continuous BLE activity takes at most 620ms.


LuDuda:
@tomchy have you tried measure switching times with the new parameters? With the new SoftDevice 6.1.1 we had to reduce timeout from 6400 to 4500 to reduce switching times, since our timeslot has been allocated exactly at timeout time (not before, even if there was no BLE activity there).

I wouldn't merge it before BLE->15.4 and 15.4->BLE timings are measured in BLE advertising, connection and scanning scenarios.

hubertmis:
I've been writing the same commend @LuDuda just published :) Or at least similar one

tomchy:
1. I thought that making a PR will trigger at least a basic verification, to see if that breaks some tests, that are already implemented.

2. Do you have any test/tools to measure those values? I can confirm from logic analyzer that a timeslot is given faster than after 620ms, but I think we should have a much more solid proof for that (with exact switching time values).

hubertmis:
@tomchy , we plan to create automatic switching time test. Right now we have to measure it manually. I suggest F2F discussion after lunch.

hubertmis:
Decided F2F that we are not going to merge this workaround. According to measurements, it decreases multi-protocol performance during protocol switching.
There is function `nrf_raal_softdevice_config` available that allows application to apply this kind of workaround whenever needed.

PR #190, opened 2019-01-25
Add new chip revisions

hubertmis:


hubertmis:
I've just added revisions to FEM configuration as well.

PR #191, opened 2019-01-25
Allow CSMA-CA to wait for an idle timeslot before CCA.

hubertmis:
This is an optional workaround for MAC layers that does not handle
correctly multiple CHANNEL_BUSY reports in a row.


hubertmis:
@wbober @tomchy 
Have a look at this new configuration parameter. It's going to be included in incoming release.

src/nrf_802154_config.h
```diff
@@ -422,6 +422,22 @@ extern "C" {
 #define NRF_802154_CSMA_CA_MAX_CSMA_BACKOFFS 4
 #endif
 
+/**
+ * @def NRF_802154_CSMA_CA_WAIT_FOR_TIMESLOT
+ *
+ * If the CSMA-CA algorithm waits for an available timeslot before it performs the CCA procedure.
```
b-gent:
``` suggestion
 * Indicates whether the CSMA-CA algorithm waits for an available timeslot before it performs the CCA procedure.
```

src/nrf_802154_config.h
```diff
@@ -422,6 +422,22 @@ extern "C" {
...
+ *
+ * @note When this option is enabled, the CSMA-CA procedure may be synchronized with timeslots of
+ *       other protocols. It decreases robustness of the CSMA-CA procedure and is not recommended.
+ *       This option may be used as a workaround to problems with higher layers that does not handle
```
b-gent:
``` suggestion
 *       This option may be used as a workaround to problems with higher layers that do not correctly handle
```

src/nrf_802154_config.h
```diff
@@ -422,6 +422,22 @@ extern "C" {
...
+ *       correctly multiple CHANNEL_BUSY errors in a row. This potentially problematic situation may
```
b-gent:
``` suggestion
 *       multiple CHANNEL_BUSY errors in a row. This potentially problematic situation may
```

PR #203, opened 2019-02-22
Fix build dependencies and list of build files

hubertmis:


e-rk:
Just my 2 cents here - one file is missing from direct variant: `nrf_802154_frame_parser.c`.

hubertmis:
@e-rk, right! Hopefully tests will fail this time :)

hubertmis:
Regression test

PR #204, opened 2019-02-23
Clean up Travis.yml

EwoutH:
- Remove license info (duplicate of LICENSE)
- Update OS to Ubuntu 1604 (xenial)
- Move Style check into its own job (this allows more jobs to be added in the future, with an own name, install, script etc..)
- Split up combined commands (with adding - and removing trailing spaces + &&)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,31 @@
+# Copyright (c) 2018-2019, Nordic Semiconductor ASA	
+# All rights reserved.	
+#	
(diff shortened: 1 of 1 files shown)


hubertmis:
Thanks for the contribution. I like cleanup of the travis configuration very much.
I'm checking if it is ok to remove license from header of this file. Moreover, it seems that there is a problem with continuous integration server in our office. It reports some false positives. We are working to fix this.

PR #205, opened 2019-02-28
Fix delayed RX with channel change

jciupis:
With the currently implemented logic, the core module does not notify that entering `RECEIVE` state has succeeded when state is already set to `RECEIVE`. It is misleading and causes delayed frame reception to fail, since its callback is never called. This PR fixes this and also corrects RX setup time to match logic analyzer measurements.


jciupis:
Regression test

hubertmis:
Regression test

hubertmis:
Regression test

hubertmis:
Regression test

hubertmis:
Regression test

hubertmis:
Regression test

PR #208, opened 2019-03-11
Pr/unify state of core after delayed trx operation regardless timeslot start succeeded or failed

aprochwicz:


tools/event_decoder/decoder.html
```diff
@@ -306,7 +352,7 @@ <h3 class="text-center">Sequence diagram generator for <span class="label label-
                 }
 
                 var event    = hexValue & 0xffff;
-                var func     = hexValue >> 16;
+                var func     = hexValue >>> 16;
```
hubertmis: What is `>>>` operator?

tools/event_decoder/decoder.py
```diff
@@ -0,0 +1,31 @@
+import os
+import re
+
```
hubertmis: Could you add some comment how does this script work?

src/nrf_802154_debug.h
```diff
@@ -81,39 +77,28 @@ extern "C" {
...
-#define FUNCTION_RSCH_PEND_GRANTED             0x0486UL
-#define FUNCTION_RSCH_PEND_REVOKED             0x0487UL
-#define FUNCTION_RSCH_NOTIFY_GRANTED           0x0488UL
-#define FUNCTION_RSCH_NOTIFY_REVOKED           0x0489UL
-#define FUNCTION_RSCH_NOTIFY_IF_PENDING        0x048AUL
 #define FUNCTION_RSCH_DELAYED_TIMESLOT_REQ     0x048BUL
 #define FUNCTION_RSCH_TIMER_DELAYED_PREC       0x048CUL
 #define FUNCTION_RSCH_TIMER_DELAYED_START      0x048DUL
 
+#define FUNCTION_mutex_trylock                            0x4900UL
```
hubertmis: Why value space `0x49xx` was selected? And why is it placed here in this file? It is out of value order.

src/rsch/nrf_802154_rsch.h
```diff
@@ -156,8 +156,7 @@ bool nrf_802154_rsch_timeslot_request(uint32_t length_us);
  *
  * Request timeslot that should be granted in the future. Function parameters provides data when
  * the timeslot should start and how long should it last. When requested timeslot starts the
- * @ref nrf_802154_rsch_delayed_timeslot_started is called. If requested timeslot cannot be granted
- * with requested parameters, the @ref nrf_802154_rsch_delayed_timeslot_failed is called.
+ * @ref nrf_802154_rsch_delayed_timeslot_started is called.
```
hubertmis: Should we add a note here that the function `nrf_802154_rsch_delayed_timeslot_started` may be delayed it is up to the called module to verify if this delay is problematic?

tools/event_decoder/decoder.html
aprochwicz: In js `>>` is a sign-propagating right shift. `>>` caused unsigned numbers > 0x7FFF to be showed as negative.
`>>>` is zero-fill right shift and all unsigned numbers are showed as positive.

src/nrf_802154_debug.h
hubertmis: Why not `0x10xx` or `0x80xx` or `0xF0xx`? I would choose a little more rounded number than 49 :)

src/rsch/raal/simulator/nrf_802154_debug.h
```diff
@@ -0,0 +1,74 @@
+/* Copyright (c) 2017 - 2018, Nordic Semiconductor ASA
```
jciupis:
```suggestion
/* Copyright (c) 2017 - 2019, Nordic Semiconductor ASA
```

src/rsch/raal/softdevice/nrf_802154_debug.h
```diff
@@ -0,0 +1,74 @@
+/* Copyright (c) 2017 - 2018, Nordic Semiconductor ASA
```
jciupis:
```suggestion
/* Copyright (c) 2017 - 2019, Nordic Semiconductor ASA
```

src/nrf_802154_debug.h
aprochwicz: If the script will be developed, I planned to assign value space for each file (same as now with manually assigned values) so the value spaces will not be round anyway.

hubertmis: How about using value space like 0x1xxx for any file and then 0x10xx for first file, 0x11xx for seconds one?

aprochwicz: Ok, I will change it.

tools/event_decoder/decoder.py
aprochwicz: Added

hubertmis:
Regression test

PR #210, opened 2019-03-13
Unify LQI type

### LINKED ISSUE #187: Mixed types of the LQI value
In the driver code LQI is kept either as `uint8_t` or as `int8_t`. It is a bug that causes implicit type conversions. LQI type should be unified to `uint8_t` according to the IEEE 802.15.4 standard (0x00 - 0xff).

hubertmis:
This PR fixes #187


aprochwicz:
One more `int8_t` in `nrf_802154_core.c`: static void transmitted_frame_notify(uint8_t * p_ack, int8_t power, int8_t lqi)

hubertmis:
Regression test

Issue #198, opened 2019-02-18
Missing subdirectory in the nrf_802154_core.c include path


pszkotak:
Hello,

It seems that `mac_features` subdirectory is missing in the `nrf_802154_frame_parser.h` inclusion.
https://github.com/NordicSemiconductor/nRF-IEEE-802.15.4-radio-driver/blob/652a35c8718125ae7ef615df1155c5c412780b17/src/nrf_802154_core.c#L50

Kind regards,
Piotr

hubertmis:
Fixed by #212

PR #212, opened 2019-03-20
Fix missing include subdirectories

jciupis:
This PR fixes issue [#198](https://github.com/NordicSemiconductor/nRF-IEEE-802.15.4-radio-driver/issues/198).


hubertmis:
Regression test

PR #213, opened 2019-03-22
Remove files copied from nrfx. These files should be provided separately

hubertmis:


hubertmis:
Manually run CI tests, passed. http://projecttools.nordicsemi.no/bamboo/browse/KNG-TDSM81-12
Problem with automatic status reported.

PR #214, opened 2019-03-25
Remove timer callback from lp_timer_start function

jciupis:


src/platform/lp_timer/nrf_802154_lp_timer_nodrv.c
```diff
@@ -570,6 +572,12 @@ void NRF_802154_RTC_IRQ_HANDLER(void)
     }
 
     // Handle compare match.
+    if (m_shall_fire_immediately)
```
hubertmis: Would it be ok to just call `handle_compare_match(false)` here? Instead of providing a new variable?

jciupis: It would make the `if (nrf_rtc_int_is_enabled...)` check in line 581 obsolete. I'm not sure if that's fine

hubertmis: The only problem I see that we would call unnecessary `handle_compare_match()` on overflow events. I don't think that's a big problem compared to saving one variable and making the code easier to maintain.

jciupis: That seems reasonable. I'll update the code and verify if it works as expected.

PR #216, opened 2019-03-28
Clean up FEM initialization.

hubertmis:
It was possible that when only PA or only LNA configuration is enabled,
the FEM module initializes GPIOTE and PPI for both channels. This patch
configures peripherals only for enabled channels.


src/fem/nrf_fem_control.c
```diff
@@ -122,15 +128,42 @@ static void gpiote_configure(void)
...
+            (m_nrf_fem_control_cfg.lna_cfg.active_high ?
+             (uint32_t)(&NRF_GPIOTE->TASKS_CLR[m_nrf_fem_control_cfg.lna_gpiote_ch_id]) :
+             (uint32_t)(&NRF_GPIOTE->TASKS_SET[m_nrf_fem_control_cfg.lna_gpiote_ch_id])),
+            (m_nrf_fem_control_cfg.pa_cfg.active_high ?
+             (uint32_t)(&NRF_GPIOTE->TASKS_CLR[m_nrf_fem_control_cfg.pa_gpiote_ch_id]) :
+             (uint32_t)(&NRF_GPIOTE->TASKS_SET[m_nrf_fem_control_cfg.pa_gpiote_ch_id])));
+    }
+    else if (m_nrf_fem_control_cfg.lna_cfg.enable)
+    {
+        nrf_ppi_channel_and_fork_endpoint_setup(
```
jciupis: It seems that both here and in line 155 `nrf_ppi_channel_endpoint_setup` would be enough.

aprochwicz: Alternatively also `nrf_ppi_event_endpoint_setup`, `nrf_ppi_task_endpoint_setup` and `nrf_ppi_fork_endpoint_setup` could be used. It would allow to remove repeated parts.

hubertmis: I'll go with @jciupis proposal. It seems that it would be the cleanest way.

The alternative approach would result in sth like:
```
if (pa || lna)
{
    set_event(DISABLED);
}

if (pa && lna)
{
  set_task_and_fork(pa, lna);
}
else if (pa)
{
  set_task(pa);
}
else if (lna)
{
  set_task(lna);
}
```

I think this code is a little harder to follow.

aprochwicz: Or simpler:
```
if (pa || lna)
{
    set_event(DISABLED);
}

if (pa)
{
  set_task(pa);
}

if (lna)
{
  set_task(lna);
}
```

hubertmis: Yeah, I thought about that, but I find it confusing when only fork is set, while task is not.
Moreover, current approach seems to be more compliant with the PPI refactoring (during internal review).

hubertmis:
Regression test

hubertmis:
Regression test

hubertmis:
Regression test

PR #220, opened 2019-04-10
Pr/fix delayed operations called in sleep state

jciupis:
This PR attempts to fix issues that occur when delayed operations are called with radio in sleep state.


src/rsch/nrf_802154_rsch.c
```diff
@@ -365,7 +348,6 @@ void nrf_802154_rsch_continuous_mode_priority_set(rsch_prio_t prio)
     __DMB();
 
     all_prec_update();
-    notify_core();
 
     if (prio == RSCH_PRIO_IDLE)
```
hubertmis: I think this `if` block should be removed now. `m_last_notified_prio` is updated by `all_prec_update()`

src/rsch/nrf_802154_rsch.c
```diff
@@ -10,7 +10,7 @@
 #include "raal/nrf_raal_api.h"
 #include "timer_scheduler/nrf_802154_timer_sched.h"
 
-#define PREC_RAMP_UP_TIME 300                                ///< Ramp-up time of preconditions [us]. 300 is worst case for HFclock
+#define PREC_RAMP_UP_TIME 400                                ///< Ramp-up time of preconditions [us]. 400 is worst case for HFclock
```
hubertmis: Is that really possible that HF clock ramps up longer than 300 us?

jciupis: While 400us might be too much and will be tested further, values over 350us were observed with trace.

hubertmis: Interesting. That's good we have some measurement results and current value looks reasonable.
Anyway, I think we should discuss it with HW engineers. If I remember correctly, product specification defines max as ~250.

jciupis: Product specification states that:

```
When the HFCLKSTART task is triggered, the HFCLKSTARTED event is generated once the
HFXO startup time has elapsed.
```

The startup time is a sum of HFXO power-up time and HFXO debounce time. Depending on the crystal used, the former varies between 50us and 210us, while the latter can be configured to a value ranging from 16us up to 4096us with recommended values for different crystals somewhere in between. Since the `PREC_RAMP_UP_TIME` affects the delayed operations feature significantly, I agree it's worth detailed investigation and discussion first.

jciupis:
This PR got a bit messy and the implementation changed significantly, so I'm closing this one and opening a new one.

PR #237, opened 2019-05-22
Add nRF52811 to supported SoCs list in README.md

### LINKED ISSUE #235: porting for the nrf52811
Is there any work in progress for porting the driver to the nRF52811?

We are currently working on the nRF52840 DK but would like to move to our own hardware soon which is using the nRF52811.

hubertmis:
Fixes #235


hubertmis:
Regression test

PR #238, opened 2019-05-22
Add random number abstraction layer in platform directory

hubertmis:
Some systems provide other pseudo-random number generator than the
standard library. In order to support these generators new platform
abstraction layer is provided for pseudo-random number generator.


KaSroka:
Great, looks good!

hubertmis:
Regression test

hubertmis:
Regression test

PR #241, opened 2019-05-27
Add possibility to update Ack IEs for an already registered address

jciupis:
Currently function `nrf_802154_ack_data_for_addr_set()` cannot be used to update Ack IEs for an already registered address. This patch changes this behaviour so that Ack IEs can be set both for registered and previously unregistered addresses.


src/mac_features/ack_generator/nrf_802154_ack_data.c
```diff
@@ -494,21 +494,36 @@ bool nrf_802154_ack_data_for_addr_set(const uint8_t * p_addr,
...
     uint32_t location = 0;
+    bool     result   = false;
 
-    // Provided address already in the database
-    if (addr_index_find(p_addr, &location, data_type, extended))
+    do
     {
-        return true;
-    }
+        if (data_type != NRF_802154_ACK_DATA_IE)
```
hubertmis: Let's imagine that we call this function to set the pending bit in Acks received from the given address. I can't find any call to `addr_add()` when we set an address for the pending bit. How does it work?

jciupis: I retained the functionality of the existing code. I thought that pending bits are entirely handled in `nrf_802154_ack_pending_bit.c`, there's a function called `nrf_802154_ack_pending_bit_for_addr_set` there (it also doesn't allow for a pending bit update btw.)

hubertmis: Right, I was convinced that this module handles both :)

src/mac_features/ack_generator/nrf_802154_ack_data.c
```diff
@@ -494,21 +494,36 @@ bool nrf_802154_ack_data_for_addr_set(const uint8_t * p_addr,
...
+        {
+            break;
```
hubertmis: As this module supports right now only DATA_IE, not pending bit, I would replace this `break` with `assert(false)`

src/mac_features/ack_generator/nrf_802154_ack_data.c
```diff
@@ -494,21 +494,36 @@ bool nrf_802154_ack_data_for_addr_set(const uint8_t * p_addr,
...
+        }
 
-    bool retval = addr_add(p_addr, location, data_type, extended);
+        // Provided address already in the database
+        if (addr_index_find(p_addr, &location, data_type, extended))
+        {
+            ie_data_add(location, extended, p_data, data_len);
+        }
+        // Provided address is unknown
+        else if (addr_add(p_addr, location, data_type, extended))
```
hubertmis: It looks that this `else if` can be safely merged with `if` above

src/mac_features/ack_generator/nrf_802154_ack_data.c
```diff
@@ -494,21 +494,32 @@ bool nrf_802154_ack_data_for_addr_set(const uint8_t * p_addr,
...
+        // Provided address is in the database already or it can be added to the database
+        if (addr_index_find(p_addr, &location, data_type, extended) ||
+            addr_add(p_addr, location, data_type, extended))
+        {
+            ie_data_add(location, extended, p_data, data_len);
+        }
+        // Provided address cannot be added
+        else
+        {
+            break;
```
hubertmis: Now this `do - while` looks strange. It would be easier to have `if () result = true; else result = false;` instead

PR #243, opened 2019-05-28
Handle pending bit setting in the ack_data module

jciupis:
This is going to be quite a big change, so I decided it's worth putting it up for review before opening a pull request.


src/nrf_802154.c
```diff
@@ -545,31 +544,31 @@ void nrf_802154_pan_coord_set(bool enabled)
 
 bool nrf_802154_ack_data_set(const uint8_t * p_addr,
                              bool            extended,
+                             uint8_t         data_type,
```
hubertmis: In this file we should not modify order of arguments. It's API breaker. Let's keep it as it is for now and create a task for version 2.0 to clean this up.

jciupis: Absolutely, thanks for noticing

hubertmis:
Regression test

PR #244, opened 2019-06-04
Fix delayed operations

jciupis:
In this PR, several issues with delayed operations scheduled when radio is asleep are fixed: processing notifications from radio scheduler, changing channels, various constants etc.


src/mac_features/nrf_802154_delayed_trx.c
```diff
@@ -50,8 +50,12 @@
 #include "rsch/nrf_802154_rsch.h"
 #include "timer_scheduler/nrf_802154_timer_sched.h"
 
-#define TX_SETUP_TIME 190u ///< Time [us] needed to change channel, stop rx and setup tx procedure.
-#define RX_SETUP_TIME 93u  ///< Time [us] needed to change channel, stop tx and setup rx procedure.
+/* The following time is the sum of 70us RTC_IRQHandler processing time, 40us of time that elapses
```
ankuns: In the nrf_802154_rsch.c line 17 there is "60us of RTC_IRQHandler processing" and in line 20 also, whereas here is 70us, Which is right? Please align.

src/rsch/nrf_802154_rsch.c
```diff
@@ -10,7 +10,21 @@
...
+ * In case of nRF52840, the value of this macro is the sum of 256us of HFXO debounce time,
+ * 75us of the worst case power-up time for an Epson crystal, 31us of timer granularity margin,
+ * 50us of POWER_CLOCK_IRQHandler processing time, 60us of RTC_IRQHandler processing time
+ * and 8us of margin.
+ * In case of nRF52811, the value of this macro is the sum of 360us of HFXO startup time,
+ * 31us of timer granularity margin, 55us of POWER_CLOCK_IRQHandler processing time, 60us of
+ * RTC_IRQHandler processing time and 9us of margin.
+ */
+#ifdef NRF52811_XXAA
+#define PREC_RAMP_UP_TIME 515                                ///< Ramp-up time of preconditions [us].
```
ankuns: Could it be defined as
```c
#define PREC_RAMP_UP_TIME (HFXO_DEVOUNCE_TIME + \
    CRYSTAL_POWER_UP_WORST_TIME + \
    TIMER_GRANULARITY_MARGIN + \
    POWER_CLOCK_IRQ_HANDLER_PROCESSING_TIME + \
    RTC_IRQHANDLER_PROCESSING_TIME)
```
And the components be defined separately. It would be easier to change individual components than calculating the sum by developer every time a component change.
The drawback is that certain number is not visible as easy as now. What do you think?

src/mac_features/nrf_802154_delayed_trx.c
jciupis: Actually both are correct ;) there are two RTC interrupts that need to be handled in order to start a delayed timeslot. First one calls `delayed_timeslot_prec_request` and that's the one specified in `nrf_802154_rsch.c`. The other one occurs later and it calls `delayed_timeslot_start`. According to measurements it is slightly longer than the first one.

ankuns: OK, thanks for clarification.

src/nrf_802154_core.c
```diff
@@ -3023,53 +3017,31 @@ bool nrf_802154_core_channel_update(void)
...
-                    channel_set(nrf_802154_pib_channel_get());
-                }
-
-                break;
-
-            case RADIO_STATE_SLEEP:
-            case RADIO_STATE_FALLING_ASLEEP:
-            case RADIO_STATE_ED:
-                // Don't perform any action in these states (channel will be updated on state change).
+            default:
```
hubertmis: Note that without `default` compiler checks during static analysis that all `enum` labels are handled. When you add `default` case compiler skips this verification. I'm ok to leave it as it is, because we didn't do anything in other cases anyway.

PR #246, opened 2019-06-07
Optimize extending delayed RX window

jciupis:
RX window is extended assuming that the longest possible frame is being received. This patch optimizes this behavior and extends RX window with actual length of the frame being received.


src/mac_features/nrf_802154_delayed_trx.c
```diff
@@ -225,15 +236,16 @@ static void notify_rx_timeout(void * p_context)
 
     if (dly_op_state_get(RSCH_DLY_RX) == DELAYED_TRX_OP_STATE_ONGOING)
     {
-        uint32_t now              = nrf_802154_timer_sched_time_get();
-        uint32_t max_frame_length = nrf_802154_rx_duration_get(MAX_PACKET_SIZE, true);
-        uint32_t sof_timestamp    = m_sof_timestamp;
+        uint32_t now           = nrf_802154_timer_sched_time_get();
+        uint32_t frame_length  = nrf_802154_rx_duration_get(m_dly_rx_frame.psdu_length,
+                                                            m_dly_rx_frame.ack_requested);
+        uint32_t sof_timestamp = m_dly_rx_frame.sof_timestamp;
```
hubertmis: I'm afraid there can be a race. Let's imagine this function is preempted by `nrf_802154_delayed_trx_rx_started_hook()`. What happens when `.psdu_length` is from a previous frame while `.ack_requested` is updated for a new one? Or the other way around?
I think if you set `.sof_timestamp` as last field in the structure (with memory barriers) in `nrf_802154_delayed_trx_rx_started_hook()`, it would be ok: the PSDU length or the ACK request bit wouldn't matter if the timestamp was in the past.

src/mac_features/nrf_802154_delayed_trx.c
```diff
@@ -467,10 +481,12 @@ bool nrf_802154_delayed_trx_abort(nrf_802154_term_t term_lvl, req_originator_t r
...
 }
 
-void nrf_802154_delayed_trx_rx_started_hook(void)
+void nrf_802154_delayed_trx_rx_started_hook(const uint8_t * p_frame)
 {
     if (dly_op_state_get(RSCH_DLY_RX) == DELAYED_TRX_OP_STATE_ONGOING)
     {
-        m_sof_timestamp = nrf_802154_timer_sched_time_get();
+        m_dly_rx_frame.sof_timestamp = nrf_802154_timer_sched_time_get();
+        m_dly_rx_frame.psdu_length   = p_frame[0];
```
hubertmis: Please, replace this magic number with `PHR_OFFSET`

jciupis:
Regression test

hubertmis:
Regression test

PR #256, opened 2019-07-23
Add RSSI measurement during idle RX feature

hubertmis:


src/nrf_802154_core.c
```diff
@@ -230,9 +232,26 @@ static void rx_flags_clear(void)
 #endif // !NRF_802154_DISABLE_BCC_MATCHING
 }
 
-/** Get result of last RSSI measurement.
+/** Request the RSSI measurement */
```
greg-fer:
```suggestion
/** Request the RSSI measurement. */
```

src/nrf_802154_core.c
```diff
@@ -230,9 +232,26 @@ static void rx_flags_clear(void)
...
+static void rssi_measure(void)
+{
+    m_flags.rssi_started = true;
+    nrf_radio_event_clear(NRF_RADIO_EVENT_RSSIEND);
+    nrf_radio_task_trigger(NRF_RADIO_TASK_RSSISTART);
+}
+
+/** Wait for the RSSI measurement */
```
greg-fer:
```suggestion
/** Wait for the RSSI measurement. */
```

src/nrf_802154_core.c
```diff
@@ -197,9 +198,10 @@ typedef struct
 
 #endif  // !NRF_802154_DISABLE_BCC_MATCHING
 #if NRF_802154_TX_STARTED_NOTIFY_ENABLED
-    bool tx_started : 1; ///< If requested transmission has started.
+    bool tx_started   : 1; ///< If requested transmission has started.
```
greg-fer:
```suggestion
    bool tx_started   : 1; ///< If the requested transmission has started.
```

src/nrf_802154_core.c
```diff
@@ -197,9 +198,10 @@ typedef struct
...
 
 #endif  // NRF_802154_TX_STARTED_NOTIFY_ENABLED
+    bool rssi_started : 1;
 } nrf_802154_flags_t;
 static nrf_802154_flags_t m_flags;               ///< Flags used to store current driver state.
```
greg-fer:
```suggestion
static nrf_802154_flags_t m_flags;               ///< Flags used to store the current driver state.
```

src/nrf_802154_core.c
```diff
@@ -230,9 +232,26 @@ static void rx_flags_clear(void)
...
+/** Wait for the RSSI measurement */
+static void rssi_measurement_wait(void)
+{
+    while (!nrf_radio_event_check(NRF_RADIO_EVENT_RSSIEND))
+    {
+        nrf_802154_busy_wait();
+    }
+}
+
+/** Get result of the last RSSI measurement.
```
greg-fer:
```suggestion
/** Get the result of the last RSSI measurement.
```

src/nrf_802154_core.c
```diff
@@ -1362,6 +1381,8 @@ static void rx_init(bool disabled_was_triggered)
 
     // Clear filtering flag
     rx_flags_clear();
+    // Clear RSSI measurement flag
```
greg-fer:
```suggestion
    // Clear the RSSI measurement flag.
```

src/nrf_802154_core.h
```diff
@@ -214,6 +214,18 @@ bool nrf_802154_core_channel_update(void);
  */
 bool nrf_802154_core_cca_cfg_update(void);
 
+/**
+ * @brief Notifies the Core module that next higher layer requested the RSSI measurement.
```
greg-fer:
```suggestion
 * @brief Notifies the core module that the next higher layer requested the RSSI measurement.
```

src/nrf_802154_swi.c
```diff
@@ -225,8 +227,19 @@ typedef struct
...
+        struct
+        {
+            bool * p_result; ///< RSSI measurement request result.
+        } rssi_measure;      ///< RSSI measurement request details.
+
+        struct
+        {
+            int8_t * p_rssi;                              ///< RSSI measurement result.
+            bool   * p_result;                            ///< RSSI measurement status.
+        } rssi_get;                                       ///< RSSI measurement result getter details.
```
greg-fer:
```suggestion
        } rssi_get;                                       ///< Getter details of the RSSI measurement result.
```
Long concatenation of nouns, so we must make it easier to read.

src/nrf_802154_swi.c
```diff
@@ -225,8 +227,19 @@ typedef struct
...
     } data;                                               ///< Request data depending on it's type.
```
greg-fer:
```suggestion
    } data;                                               ///< Request data depending on its type.
```
Serious punctuation issue.

src/nrf_802154_swi.c
hubertmis: Would `Details of the RSSI measurement result getter` be ok? I think it is more accurate

hubertmis: or `Details of the getter that retrieves the RSSI measurement result.`

greg-fer: The second option sounds best, I think.

src/nrf_802154_core.c
```diff
@@ -3071,6 +3094,45 @@ bool nrf_802154_core_cca_cfg_update(void)
...
+
+        nrf_802154_critical_section_exit();
+    }
+
+    return result;
+}
+
+bool nrf_802154_core_last_rssi_measurement_get(int8_t * p_rssi)
+{
+    bool result = m_flags.rssi_started && critical_section_enter_and_verify_timeslot_length();
```
ankuns: Right hand operand of `&&` operator, `critical_section_enter_and_verify_timeslot_length` may be not evaluated at all. Is this intentional?

src/nrf_802154_core.c
```diff
@@ -3071,6 +3094,45 @@ bool nrf_802154_core_cca_cfg_update(void)
     return result;
 }
 
+bool nrf_802154_core_rssi_measure(void)
+{
+    bool result = critical_section_enter_and_verify_timeslot_length();
+
+    if (result)
+    {
+        if (timeslot_is_granted() && m_state == RADIO_STATE_RX)
```
ankuns: Would be nice if you use explicit parenthesis, e.g.  `timeslot_is_granted() && (m_state == RADIO_STATE_RX)`

src/nrf_802154_core.c
```diff
@@ -3071,6 +3094,45 @@ bool nrf_802154_core_cca_cfg_update(void)
...
+
+bool nrf_802154_core_last_rssi_measurement_get(int8_t * p_rssi)
+{
+    bool result = m_flags.rssi_started && critical_section_enter_and_verify_timeslot_length();
+
+    if (result)
+    {
+        if (timeslot_is_granted())
+        {
+            rssi_measurement_wait();
```
ankuns: Another question. What if user code calls `nrf_802154_rssi_last_get` without former `nrf_802154_rssi_measure` ? Will it hang?

src/nrf_802154.h
```diff
@@ -901,8 +901,11 @@ bool nrf_802154_buffer_free_immediately(uint8_t * p_data);
  *
  * The result will be available after 8 us and can be read by
  * @ref nrf_802154_rssi_last_get.
+ *
+ * @retval true  RSSI measurement successfully requested.
+ * @retval false RSSI measurement cannot be scheduled at the moment.
  */
-void nrf_802154_rssi_measure(void);
+bool nrf_802154_rssi_measure(void);
```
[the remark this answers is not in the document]
hubertmis: That's a good point. We should not change API that's already present, but this API was never implemented -> nobody uses it. I'll change the name

src/nrf_802154.h
```diff
@@ -901,8 +901,11 @@ bool nrf_802154_buffer_free_immediately(uint8_t * p_data);
  *
  * The result will be available after 8 us and can be read by
  * @ref nrf_802154_rssi_last_get.
+ *
```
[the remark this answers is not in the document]
hubertmis: Should be, but I don't know what value would be measured. Probably the API user does not want to call `..._transmit` between these functions.

src/nrf_802154_core.c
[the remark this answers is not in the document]
hubertmis: According to the product specification it should be max 0.25 us. I think that's acceptable and it's intentional.

src/nrf_802154_swi.c
```diff
@@ -225,9 +227,20 @@ typedef struct
...
+        {
+            bool * p_result; ///< RSSI measurement request result.
+        } rssi_measure;      ///< RSSI measurement request details.
+
+        struct
+        {
+            int8_t * p_rssi;                              ///< RSSI measurement result.
+            bool   * p_result;                            ///< RSSI measurement status.
+        } rssi_get;                                       ///< Details of the getter that retrieves the RSSI measurement result.
+    } data;                                               ///< Request data depending on its type.
```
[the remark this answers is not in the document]
hubertmis: Uncrustify did it. I'll try to fix, but no promises. It has its own ways...

src/nrf_802154_core.c
hubertmis: Yes, it's intentional. Shall I split it to multiple lines to indicate that it's intentional?

src/nrf_802154_core.c
hubertmis: `m_flags.rssi_started` is checked at the beginning of the getting result function to prevent such hangs.

src/nrf_802154_core.c
ankuns: The function is so simple, but please split it. Having right hand operand of `&&` `||` performing side effects like here is error-prone.

src/nrf_802154.h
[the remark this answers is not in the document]
ankuns: What is the point of splitting the functionality into two parts then?

src/nrf_802154_core.c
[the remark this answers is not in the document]
ankuns: What about time 8us stated in dox comment for nrf_802154_rssi_measure in src/nrf_802154.h

src/nrf_802154_swi.c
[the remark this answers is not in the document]
ankuns: ok

src/nrf_802154.h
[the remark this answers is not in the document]
hubertmis: To keep the same API for other radios where RSSI measurement time may be significant. API user could do some CPU processing between these calls instead of busy wait.

src/nrf_802154_core.c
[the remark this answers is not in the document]
hubertmis: That's probably max time from the engineering sample. Honestly, I don't remember :) I'll verify if we support anything with longer RSSI measurement time

src/nrf_802154.h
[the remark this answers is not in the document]
ankuns: ok

src/nrf_802154_core.c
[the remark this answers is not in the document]
ankuns: ok, I accept

src/nrf_802154_core.c
```diff
@@ -3062,6 +3085,54 @@ bool nrf_802154_core_cca_cfg_update(void)
...
+
+bool nrf_802154_core_last_rssi_measurement_get(int8_t * p_rssi)
+{
+    bool result = m_flags.rssi_started;
+
+    if (result)
+    {
+        result = critical_section_enter_and_verify_timeslot_length();
+
+        if (result)
```
czeslawmakarski: Imho this part can be not nested. This would decrease the nesting. This would increase the readability.

hubertmis: I've renamed variables as well. I hope the code clearer now

hubertmis:
Regression test

hubertmis:
Regression test

PR #263, opened 2019-08-05
Remove possibility of uninitialized pointer access

jciupis:
Compiling with NDEBUG flag causes asserts to have no effect, which currently might lead to deferencing uninitialized pointers if unsupported Ack data type is used. This patch fixes this behaviour by returning false if unsupported data type is detected.


src/mac_features/ack_generator/nrf_802154_ack_data.c
```diff
@@ -307,10 +308,16 @@ static bool addr_index_find(const uint8_t * p_addr,
             break;
 
         default:
+            valid_data_type = false;
```
ankuns: It the assert still relevant?

hubertmis: I think it is. Isn't it?

ankuns: This is debatable issue. This function returns `false` in case of `data_type` not handled by the switch above. NDEBUG-built caller code has to handle such issue somehow.

hubertmis: Callers handle such issues anyway. API did not change.
The assertion is still valid regardless we enable it for debug session or disable for the product.

jciupis:
Regression test

PR #264, opened 2019-08-12
Remove list of supported devices from FEM header

jciupis:


jciupis:
Regression test

jciupis:
Regression test

PR #267, opened 2019-08-13
Use correct header providing RNG API from SoftDevice

hubertmis:


hubertmis:
Regression test

hubertmis:
Regression test

hubertmis:
Regression test

PR #270, opened 2019-09-10
Pr/zigbee source matching

JakubPegza:
Implement source address matching algorithm selection mechanism.

When calling function `nrf_802154_ack_data_pending_bit_should_be_set` one of several methods of determining the state of pending bit may now be used. Method is selected by calling `nrf_802154_source_addr_matching_method_set` after driver initialization, but before enabling the transceiver. Currently supported methods are:
 * Thread - based on previously used algorithm
 * ZigBee - based on implementation in ZBOSS stack
 * Always 1 - IEEE 802.15.4 standard compliant method, always sets the pending bit.
Minor changes include:
 * Moving include of nrf_802154_types.h to ack data header, instead of source file
 * Implementing unit tests for provided functionality


src/mac_features/ack_generator/nrf_802154_ack_data.h
```diff
@@ -99,6 +101,19 @@ bool nrf_802154_ack_data_for_addr_clear(const uint8_t * p_addr, bool extended, u
  */
 void nrf_802154_ack_data_reset(bool extended, uint8_t data_type);
 
+/**
+ * @brief Select the source matching algorithm.
```
greg-fer:
```suggestion
 * @brief Selects the algorithm that matches the source.
```

src/mac_features/ack_generator/nrf_802154_ack_data.h
```diff
@@ -99,6 +101,19 @@ bool nrf_802154_ack_data_for_addr_clear(const uint8_t * p_addr, bool extended, u
...
+ *
+ * @note This method should be called after driver initialization, but before transceiver is enabled.
```
greg-fer:
```suggestion
 * @note This function is to be called after the driver initialization, but before the transceiver is enabled.
```

src/mac_features/ack_generator/nrf_802154_ack_data.h
```diff
@@ -99,6 +101,19 @@ bool nrf_802154_ack_data_for_addr_clear(const uint8_t * p_addr, bool extended, u
...
+ *
+ * When calling @ref nrf_802154_ack_data_pending_bit_should_be_set method, it will choose one of
```
greg-fer:
```suggestion
 * When calling @ref nrf_802154_ack_data_pending_bit_should_be_set, this function will choose one of
```
`it` refers to this function, correct?

src/mac_features/ack_generator/nrf_802154_ack_data.h
```diff
@@ -99,6 +101,19 @@ bool nrf_802154_ack_data_for_addr_clear(const uint8_t * p_addr, bool extended, u
...
+ * several source address matching algorithms. In order for specific algorithm to be chosen, this
```
greg-fer:
```suggestion
 * several algorithms for source address matching. To ensure a specific algorithm is chosen,
```

src/mac_features/ack_generator/nrf_802154_ack_data.h
```diff
@@ -99,6 +101,19 @@ bool nrf_802154_ack_data_for_addr_clear(const uint8_t * p_addr, bool extended, u
...
+ * function should be called beforehand.
```
greg-fer:
```suggestion
 * call this function before @ref nrf_802154_ack_data_pending_bit_should_be_set.
```

src/nrf_802154.h
```diff
@@ -1002,7 +1002,30 @@ void nrf_802154_pan_coord_set(bool enabled);
 bool nrf_802154_pan_coord_get(void);
 
 /**
- * @brief Adds the address of a peer node for which the provided ACK data is to be set.
+ * @brief Select the source matching algorithm.
```
greg-fer: Mirror edits from nrf_802154_ack_data.h

src/nrf_802154.h
```diff
@@ -1002,7 +1002,30 @@ void nrf_802154_pan_coord_set(bool enabled);
...
+ *
+ * @param[in]  match_method Source address matching method to be used.
+ */
+void nrf_802154_src_addr_matching_method_set(nrf_802154_src_addr_match_t match_method);
+
+/**
+ * @brief Adds the address of a peer node for which the provided ACK data
+ * is to be added to the pending bit list.
+ *
+ * Pending bit list works differently, depending on upper layer for which the source
```
greg-fer:
```suggestion
 * The pending bit list works differently depending on the upper layer for which the source
```

src/nrf_802154.h
```diff
@@ -1002,7 +1002,30 @@ void nrf_802154_pan_coord_set(bool enabled);
...
+ * address matching method is selected.
+ * In case of Thread, pending bit is set only for the addresses found in the list.
```
greg-fer: Add blank line above this line.

greg-fer:
```suggestion
 * - For Thread, the pending bit is set only for the addresses found in the list.
```

src/nrf_802154.h
```diff
@@ -1002,7 +1002,30 @@ void nrf_802154_pan_coord_set(bool enabled);
...
+ * In case of Zigbee, pending bit is cleared only for the short addresses found in the list.
```
greg-fer:
```suggestion
 * - For Zigbee, the pending bit is cleared only for the short addresses found in the list.
```

src/nrf_802154.h
```diff
@@ -1002,7 +1002,30 @@ void nrf_802154_pan_coord_set(bool enabled);
...
+ * In case of standard-compliant method, pending bit is always true, which requires
```
greg-fer:
```suggestion
 * - For the standard-compliant method, the pending bit is always true, which requires
```
Are these exceptions also the only differences between the three methods? If so, mention it somewhere, for example in the documentation in src/nrf_802154_types.h

src/nrf_802154.h
```diff
@@ -1002,7 +1002,30 @@ void nrf_802154_pan_coord_set(bool enabled);
...
+ * empty data frame with AR set to 0 to be transmitted immediately afterwards.
+ *
+ * Used method can be set during initialization phase by calling @ref nrf_802154_src_matching_method.
```
greg-fer:
```suggestion
 * The source address matching method can be set during the initialization phase by calling @ref nrf_802154_src_matching_method.
```
`used method`? meaning `method to be used`?

src/nrf_802154.h
```diff
@@ -1020,10 +1043,19 @@ bool nrf_802154_ack_data_set(const uint8_t * p_addr,
                              uint8_t         data_type);
 
 /**
- * @brief Removes the address of a peer node for which the ACK data is set.
+ * @brief Removes the address of a peer node for which the ACK data is set from the pending bit list.
  *
  * The ACK data that was previously set for the given address is automatically removed.
  *
+ * Pending bit list works differently, depending on upper layer for which the source
```
greg-fer: Mirror edits from lines 1021-1028.

src/nrf_802154.h
```diff
@@ -1055,7 +1087,16 @@ bool nrf_802154_ack_data_clear(const uint8_t * p_addr, bool extended, uint8_t da
 void nrf_802154_auto_pending_bit_set(bool enabled);
 
 /**
- * @brief Adds address of a peer node for which there is pending data in the buffer.
+ * @brief Adds address of a peer node to the pending bit list.
```
greg-fer:
```suggestion
 * @brief Adds the address of a peer node to the pending bit list.
```

src/nrf_802154.h
```diff
@@ -1055,7 +1087,16 @@ bool nrf_802154_ack_data_clear(const uint8_t * p_addr, bool extended, uint8_t da
 void nrf_802154_auto_pending_bit_set(bool enabled);
 
 /**
- * @brief Adds address of a peer node for which there is pending data in the buffer.
+ * @brief Adds address of a peer node to the pending bit list.
+ *
+ * Pending bit list works differently, depending on upper layer for which the source
```
greg-fer: Mirror edits from lines 1021-1028.

src/nrf_802154.h
```diff
@@ -1068,7 +1109,16 @@ void nrf_802154_auto_pending_bit_set(bool enabled);
 bool nrf_802154_pending_bit_for_addr_set(const uint8_t * p_addr, bool extended);
 
 /**
- * @brief Removes address of a peer node for which there is no more pending data in the buffer.
+ * @brief Removes address of a peer node from the pending bit list.
+ *
+ * Pending bit list works differently, depending on upper layer for which the source
```
greg-fer: Mirror edits from lines 1021-1028.

src/nrf_802154.h
```diff
@@ -1081,6 +1131,15 @@ bool nrf_802154_pending_bit_for_addr_clear(const uint8_t * p_addr, bool extended
 /**
  * @brief Removes all addresses of a given type from the pending bit list.
  *
+ * Pending bit list works differently, depending on upper layer for which the source
```
greg-fer: Mirror edits from lines 1021-1028.

src/nrf_802154_const.h
```diff
@@ -88,6 +88,16 @@
 #define KEY_ID_MODE_2                0x10                                         ///< Bits containing the 0x10 Key Identifier Mode.
 #define KEY_ID_MODE_3                0x18                                         ///< Bits containing the 0x11 Key Identifier Mode.
 
+#define MAC_CMD_ASSOC_REQ            0x01                                         ///< MAC Association request command frame identifier.
```
greg-fer:
```suggestion
#define MAC_CMD_ASSOC_REQ            0x01                                         ///< Command frame identifier for MAC Association request.
```
This is quite a long concatenation of nouns... so please rework all the defines in lines 91-99 as this one. Alternative is `Command frame identifier: MAC Association request.`

src/nrf_802154_types.h
```diff
@@ -138,6 +138,15 @@ typedef uint8_t nrf_802154_ack_data_t;
 #define NRF_802154_ACK_DATA_PENDING_BIT 0x00
 #define NRF_802154_ACK_DATA_IE          0x01
 
+/**
+ * @brief Methods of source address matching.
+ */
+typedef uint8_t nrf_802154_src_addr_match_t;
+
+#define NRF_802154_SRC_ADDR_MATCH_THREAD   0x00 // !< Implementation for Thread protocol.
+#define NRF_802154_SRC_ADDR_MATCH_ZIGBEE   0x01 // !< Implementation for ZigBee protocol.
```
greg-fer:
```suggestion
#define NRF_802154_SRC_ADDR_MATCH_ZIGBEE   0x01 // !< Implementation for the Zigbee protocol.
```
We write `ZigBee` as `Zigbee` at Nordic.
Plus, see question about the Thread protocol in the previous line.

src/nrf_802154_types.h
```diff
@@ -138,6 +138,15 @@ typedef uint8_t nrf_802154_ack_data_t;
 #define NRF_802154_ACK_DATA_PENDING_BIT 0x00
 #define NRF_802154_ACK_DATA_IE          0x01
 
+/**
+ * @brief Methods of source address matching.
+ */
+typedef uint8_t nrf_802154_src_addr_match_t;
+
+#define NRF_802154_SRC_ADDR_MATCH_THREAD   0x00 // !< Implementation for Thread protocol.
```
greg-fer:
```suggestion
#define NRF_802154_SRC_ADDR_MATCH_THREAD   0x00 // !< Implementation for the Thread protocol.
```
Are we talking here about Nordic's protocol? Then why not use `for Nordic's Thread protocol`? @hubertmis ?

src/nrf_802154_types.h
```diff
@@ -138,6 +138,15 @@ typedef uint8_t nrf_802154_ack_data_t;
...
+#define NRF_802154_SRC_ADDR_MATCH_ZIGBEE   0x01 // !< Implementation for ZigBee protocol.
+#define NRF_802154_SRC_ADDR_MATCH_ALWAYS_1 0x02 // !< Standard-compliant implementation - Pending bit is always set to 1.
```
greg-fer:
```suggestion
#define NRF_802154_SRC_ADDR_MATCH_ALWAYS_1 0x02 // !< Standard-compliant implementation. The pending bit is always set to 1.
```

src/nrf_802154.h
greg-fer: Would it make sense to move the following lines here and in other functions to the nrf_802154_src_addr_match_t typedef documentation?
```
 * In case of Thread, pending bit is set only for the addresses found in the list.
 * In case of Zigbee, pending bit is cleared only for the short addresses found in the list.
 * In case of standard-compliant method, pending bit is always true, which requires
 * empty data frame with AR set to 0 to be transmitted immediately afterwards.
```
Here and in all the functions that follow, you could link to the three methods:
```
- For Thread, @ref NRF_802154_SRC_ADDR_MATCH_THREAD
...
For details, see @ref nrf_802154_src_addr_match_t.
```
You save some space and make the documentation less prone to copypasting errors.

src/nrf_802154_types.h
hubertmis: No, it's not Nordic's protocol. It's common specification for all Thread Group members.

src/mac_features/ack_generator/nrf_802154_ack_data.h
JakubPegza: I'm not sure about the wording here. This sounds to me like the algorithm is supposed to be matched to the source, while this should mean that the algorithm is performing the source matching.

greg-fer: Ignore this suggestion then.

src/mac_features/ack_generator/nrf_802154_ack_data.h
greg-fer: Sounds ok.

test/unit tests/ack_source_matching/unity_test.c
```diff
@@ -0,0 +1,334 @@
...
+    // Extended
+    test_extended_flag = true;
+
+    nrf_802154_frame_parser_mhr_parse_ExpectAndReturn(test_psdu_extended, &test_mhr_data_extended, true);
+    nrf_802154_frame_parser_mhr_parse_IgnoreArg_p_fields();
+    nrf_802154_frame_parser_mhr_parse_ReturnThruPtr_p_fields(&test_mhr_data_extended);
+
+    res = nrf_802154_ack_data_pending_bit_should_be_set(test_psdu_extended);
+    TEST_ASSERT_TRUE(res);
+}
```
jciupis: Please add an empty line at the end of the file

src/nrf_802154.h
```diff
@@ -1002,7 +1002,31 @@ void nrf_802154_pan_coord_set(bool enabled);
...
+ * @param[in]  match_method Source address matching method to be used.
+ */
+void nrf_802154_src_addr_matching_method_set(nrf_802154_src_addr_match_t match_method);
+
+/**
+ * @brief Adds the address of a peer node for which the provided ACK data
+ * is to be added to the pending bit list.
+ *
+ * The pending bit list works differently, depending on the upper layer for which the source
+ * address matching method is selected.
```
greg-fer:
```suggestion
* address matching method is selected:
```
Mirror the edit elsewhere in the PR.

src/nrf_802154.h
```diff
@@ -1002,7 +1002,31 @@ void nrf_802154_pan_coord_set(bool enabled);
...
+ *
```
greg-fer: Remove this blank line here and elsewhere in the PR.

src/nrf_802154.h
```diff
@@ -1002,7 +1002,31 @@ void nrf_802154_pan_coord_set(bool enabled);
...
+ * For Thread, @ref NRF_802154_SRC_ADDR_MATCH_THREAD
```
greg-fer: Lines 1024-1026 here and elsewhere need to be bullet points, as initially suggested. Keeping them like this now will render these lines as 1 paragraph.

src/nrf_802154.h
```diff
@@ -1020,10 +1044,19 @@ bool nrf_802154_ack_data_set(const uint8_t * p_addr,
...
  * The ACK data that was previously set for the given address is automatically removed.
  *
+ * The pending bit list works differently, depending on the upper layer for which the source
+ * address matching method is selected.
+ *
+ * For Thread, @ref NRF_802154_SRC_ADDR_MATCH_THREAD
+ * For Zigbee, @ref NRF_802154_SRC_ADDR_MATCH_ZIGBEE
+ * For Standard-compliant, @ref NRF_802154_SRC_ADDR_MATCH_ALWAYS_1
+ *
+ * Used method can be set during initialization phase by calling @ref nrf_802154_src_matching_method.
```
greg-fer:
```suggestion
* The method can be set during initialization phase by calling @ref nrf_802154_src_matching_method.
```
Mirror elsewhere in the file.

src/nrf_802154_types.h
```diff
@@ -138,6 +138,19 @@ typedef uint8_t nrf_802154_ack_data_t;
 #define NRF_802154_ACK_DATA_PENDING_BIT 0x00
 #define NRF_802154_ACK_DATA_IE          0x01
 
+/**
+ * @brief Methods of source address matching.
+ */
+typedef uint8_t nrf_802154_src_addr_match_t;
+
+#define NRF_802154_SRC_ADDR_MATCH_THREAD   0x00 // !< Implementation for the Thread protocol.
+                                                // !< The pending bit is set only for the addresses found in the list.
```
greg-fer: We don't add documentation lines like this. Move this information to the typedef documentation above, and add refs to the defines.
```
@brief Methods of source address matching.

You can use one of the following methods that can be set during the initialization phase
by calling @ref nrf_802154_src_matching_method:
- For Thread: @ref NRF_802154_SRC_ADDR_MATCH_THREAD -- The pending bit is set only for the addresses found in the list.
- For Zigbee: @ref NRF_802154_SRC_ADDR_MATCH_ZIGBEE -- The pending bit is cleared only for the short addresses found in the list. This method does not set pending bit in non-command and non-data-request frames.
- For standard-compliant implementation: @ref NRF_802154_SRC_ADDR_MATCH_ALWAYS_1 -- The pending bit is always set to 1. This requires an empty data frame with AR set to 0 to be transmitted immediately afterwards.
```
Also, you haven't added the link to the typedef in nrf_802154.h's functions. Was this intended? We need it to mention the typedef, as we currently don't mention that there are important notes for each define.

src/nrf_802154.h
```diff
@@ -1081,6 +1132,15 @@ bool nrf_802154_pending_bit_for_addr_clear(const uint8_t * p_addr, bool extended
 /**
  * @brief Removes all addresses of a given type from the pending bit list.
  *
+ * The pending bit list works differently, depending on the upper layer for which the source
```
greg-fer: See previous comments.

JakubPegza:
> The source matching algorithm for data.requests should be like following:
> 
> If (DATA REQUEST source address is the address of the known device)
> {
> If (We have indirect data for this address)
> {
> Set pending bit to 1
> }
> else
> {
> Set pending bit to 0
> }
> }
> else
> {
> Set pending bit to 1 //unknown or aged device
> }
> 
> The problem with your implementation is that if the device is in the address table (first 'if') but there is no indirect data (second 'if') for this device the pending bit will be still set and that's not correct. I see two potential solutions: first is to use the other address table for neighbors and the second is to add 'neighbor' and 'indirrect_data' flags to the existing pending table entries and modify nrf_802154_pending_bit_for_addr_set/clear access functions.

@DSRTechWriter I'm not sure if that's entirely correct. This implementation changes the way the address table works, as documented in header files. It's based on the fact that there is only one case in which the pending bit is not set in response to data requests - source address is known and no indirect data is present - as specified by you. 

This way, the higher layer needs to append all known addresses for which we do not have data to the pending bit list - as opposed to previous (Thread) implementation, where the higher layer needed to append all known addresses for which we do have data. Those two ifs in your algorithm are merged into one - we don't check whether the address is known and whether we have data for it separately.  Instead, single check is made to determine whether the address is known and we don't have data for it.

This approach simplifies the implementation by not requiring separate tracking of the known addresses and pending data by the driver, although the usage of the pending bit list by the higher layer is slightly different. Unless there is something else I have missed, in which case please correct me.

JakubPegza:
Regression test

PR #280, opened 2019-10-14
pr/rx_wifi_coex_pib_config Added Coex rx request modes to pib and public API.

ankuns:
Not sure if target should be `int/wifi`.

### DIFF
--- a/src/nrf_802154.c
+++ b/src/nrf_802154.c
@@ -161,35 +161,9 @@ int8_t nrf_802154_tx_power_get(void)
     return nrf_802154_pib_tx_power_get();
 }
 
-static bool is_rx_request_mode_supported(nrf_802154_coex_rx_request_mode_t mode)
-{
-    bool result = false;
-    switch (mode)
-    {
-        case NRF_802154_COEX_RX_REQUEST_MODE_DISABLED:
-        case NRF_802154_COEX_RX_REQUEST_MODE_ENERGY_DETECTION:
-        case NRF_802154_COEX_RX_REQUEST_MODE_PREAMBLE:
-        case NRF_802154_COEX_RX_REQUEST_MODE_DESTINED:
-            result = true;
-            break;
-
-        default:
-            break;
-    }
-    return result;
-}
-
 bool nrf_802154_coex_rx_request_mode_set(nrf_802154_coex_rx_request_mode_t mode)
 {
-    bool result = false;
-
-    if (is_rx_request_mode_supported(mode))
-    {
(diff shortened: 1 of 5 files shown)


src/nrf_802154_types.h
```diff
@@ -161,6 +161,22 @@ typedef uint8_t nrf_802154_src_addr_match_t;
...
+ *
+ * Possible values:
+ * - @ref NRF_802154_COEX_RX_REQUEST_MODE_DISABLED,
+ * - @ref NRF_802154_COEX_RX_REQUEST_MODE_ENERGY_DETECTION,
+ * - @ref NRF_802154_COEX_RX_REQUEST_MODE_PREAMBLE,
+ * - @ref NRF_802154_COEX_RX_REQUEST_MODE_DESTINED
+ */
+typedef uint8_t nrf_802154_coex_rx_request_mode_t;
+
+#define NRF_802154_COEX_RX_REQUEST_MODE_DISABLED           0x00    //!< Coex does not request to arbiter in receive mode.
```
czeslawmakarski: Why not enum?

ankuns: This way fits into already used convention.

hubertmis: You never know size of enum in C. It's not safe to use enum in public API definitions

src/nrf_802154_pib.c
```diff
@@ -57,6 +57,7 @@ typedef struct
     bool                 auto_ack    : 1;                      ///< Indicating if auto ACK procedure is enabled.
     bool                 pan_coord   : 1;                      ///< Indicating if radio is configured as the PAN coordinator.
     uint8_t              channel     : 5;                      ///< Channel on which the node receives messages.
+    nrf_802154_coex_rx_request_mode_t coex_rx_request_mode;    ///< Coex request mode for received frames.
```
jciupis: I hate to say this, but shouldn't it all be indented to match the coex field?

src/nrf_802154.c
```diff
@@ -161,6 +161,42 @@ int8_t nrf_802154_tx_power_get(void)
     return nrf_802154_pib_tx_power_get();
 }
 
+static bool is_rx_request_mode_supported(nrf_802154_coex_rx_request_mode_t mode)
```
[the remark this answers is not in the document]
ankuns: so let's move this into pib

src/nrf_802154_pib.c
hubertmis: I didn't want to say it loud. I hoped Travis would do that instead :)

PR #284, opened 2019-10-15
Tweaked GPIO Debug for internal usage; Restored define.

czeslawmakarski:
- Removed nrf_802154_debug.c from .packsc
- Restored original define.


nRF-IEEE-802.15.4-radio-driver.packsc
```diff
@@ -45,6 +45,8 @@
                     "src/nrf_802154_core.c",
                     "src/nrf_802154_core_hooks.c",
                     "src/nrf_802154_critical_section.c",
+                    "src/nrf_802154_debug_log.c",
```
ankuns: Shouldn't those files be included somewhere around lines 440 with "debug_log" and "debug_assert"? Mots of th content of these files are compiled conditionally with option turned on by "debug_log" "debug_assert"

src/nrf_802154_debug_assert.c
```diff
@@ -0,0 +1,75 @@
+/* Copyright (c) 2017 - 2018, Nordic Semiconductor ASA
```
hubertmis: nit: 2019?

PR #285, opened 2019-10-17
Pr/coex_rx_config_trx

ankuns:
Implemented functionality in trx providing `nrf_802154_trx_receive_frame_prestarted` and `nrf_802154_trx_receive_frame_started` callouts.

Due to no direct interrupt generated by the RADIO peripheral, "prestarted" event is generated through ppi and egu/swi. PPI channel has the same number as the one used in NRF_802154_DISABLE_BCC_MATCHING=1 mode, thus "prestarted" functionality is available only when NRF_802154_DISABLE_BCC_MATCHING=0.

This required reworking of nrf_802154_swi.c file which is now included in all variants.
"notification"-related functionality moved from nrf_802154_swi.c to nrf_802154_notification_swi.c
"request"-related functionality moved from nrf_802154_swi.c to nrf_802154_request_swi.c
"hfclk"-related functionality moved from nrf_802154_swi.c to nrf_802154_priority_drop_swi.c

Above shared common "queue" functionality which is provided in a new file nrf_802154_queue.c (.h)

In the PIB only those modes can be accepted by a call to nrf_802154_pib_coex_rx_request_mode_set which are really supported.


src/nrf_802154_trx.h
```diff
@@ -340,6 +361,47 @@ void nrf_802154_trx_abort(void);
...
+ * @note This handler may be triggered several times during receive of a preamble
+ *       of a frame. It can be followed by call to @ref nrf_802154_trx_receive_frame_started
+ *       (if enabled) and then by @ref nrf_802154_trx_receive_frame_crcok or
+ *       @ref nrf_802154_trx_receive_frame_crcerror, but there is possibility
+ *       that it will not be followed by these calls (In case when the RADIO didn't found
+ *       full preamble.). If implementation of this handler starts some
+ *       activity that must be terminated by a further call, it must implement
+ *       its own timeout feature.
+ */
+extern void nrf_802154_trx_receive_frame_prestarted(void); // TODO: is this name meaningful?
```
jciupis: This `TODO` comment should probably be removed

src/nrf_802154_trx.c
```diff
@@ -70,16 +76,22 @@
                                        (1U << PPI_ADDRESS_COUNTER_COUNT) | \
                                        (1U << PPI_CRCERROR_COUNTER_CLEAR))
 #else
-#define PPI_NO_BCC_MATCHING_USED_MASK 0U
-#endif // NRF_802154_DISABLE_BCC_MATCHING
+#define PPI_RADIO_HELPER1_EGU_HELPER1 NRF_802154_PPI_RADIO_HELPER1_TO_EGU_HELPER1 ///!< PPI that connects RADIO HELPER1 event with EGU task for HELPER1 channel
+
+#define PPI_NO_BCC_MATCHING_USED_MASK (1 << PPI_RADIO_HELPER1_EGU_HELPER1)
```
jciupis: Why not use the mask macro defined in the peripherals header file? It already has this value

src/nrf_802154_queue.c
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * @brief Module implementing simple FIFO queue.
+ */
+
+#include <assert.h>
+
+#include "nrf_802154_queue.h"
+
+static inline uint8_t increment_modulo(uint8_t v, uint8_t wrap_at_value)
```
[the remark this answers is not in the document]
ankuns: ... I partially agree, but `max_value` is flawed in the same way. Is `max_value` the maximum legal value or should it be not reached.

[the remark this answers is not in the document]
ankuns: I would leave it as is if you don't mind.

src/nrf_802154_trx.c
ankuns: This is because PPI_NO_BCC_MATCHING_USED_MASK applies to trx only, and NRF_802154_DISABLE_BCC_MATCHING_PPI_CHANNELS_USED_MASK applies to the whole radio driver.

jciupis: Makes sense to leave it as it is then :+1:

src/nrf_802154_request_swi.c
```diff
@@ -99,14 +250,239 @@ static bool active_vector_priority_is_high(void)
...
+    req_exit();
+}
+
+/**
+ * @brief Gets the last RSSI measurement result from the core module.
+ *
+ * @param[out] p_result Pointer where the result to be returned by
+ *                      nrf_802154_request_rssi_measurement_get should be written by the swi handler.
+ */
+static void swi_rssi_measurement_get(int8_t * p_rssi, bool * p_result)
```
JakubPegza: Undocumented `p_rssi` argument

ankuns: done

PR #291, opened 2019-10-23
Pr/trx tx notifications

czeslawmakarski:
The implementation of TRX CCA Idle callout.

### DIFF
--- a/src/nrf_802154_core.c
+++ b/src/nrf_802154_core.c
@@ -741,7 +741,15 @@ static bool tx_init(const uint8_t * p_data, bool cca, bool disabled_was_triggere
         return false;
     }
 
-    nrf_802154_trx_transmit_frame(p_data, cca);
+    nrf_802154_trx_receive_notifications_t notifications_mask = TRX_TRANSMIT_NOTIFICATION_NONE;
+    if (nrf_802154_coex_tx_request_mode_get() == NRF_802154_COEX_TX_REQUEST_CCA_DONE)
+    {
+        notifications_mask |= TRX_TRANSMIT_NOTIFICATION_CCAIDLE;
+    }
+
+    nrf_802154_trx_transmit_frame(p_data,
+                                  cca,
+                                  notifications_mask);
 
     return true;
 }
@@ -1466,6 +1474,11 @@ void nrf_802154_trx_standalone_cca_finished(bool channel_was_idle)
     rx_init(true);
 
     cca_notify(channel_was_idle);
+}
+
+void nrf_802154_trx_transmit_frame_ccaidle(void)
+{
+    // Intentionally empty
 }
 
 void nrf_802154_trx_transmit_frame_ccabusy(void)
--- a/src/nrf_802154_core.c
+++ b/src/nrf_802154_core.c
@@ -120,6 +120,8 @@ static volatile rsch_prio_t m_rsch_priority = RSCH_PRIO_IDLE; ///< Last notified
 
 /** @brief Value of argument @c notifications_mask to @ref nrf_802154_trx_receive_frame */
 static nrf_802154_trx_receive_notifications_t m_trx_receive_frame_notifications_mask;
+/** @brief Value of argument @c notifications_mask to @ref nrf_802154_trx_transmit_frame */
+static nrf_802154_trx_transmit_notifications_t m_trx_transmit_frame_notifications_mask;
 
 /***************************************************************************************************
  * @section Common core operations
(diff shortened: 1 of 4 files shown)


src/nrf_802154_core.c
```diff
@@ -741,7 +741,15 @@ static bool tx_init(const uint8_t * p_data, bool cca, bool disabled_was_triggere
         return false;
     }
 
-    nrf_802154_trx_transmit_frame(p_data, cca);
+    nrf_802154_trx_receive_notifications_t notifications_mask = TRX_TRANSMIT_NOTIFICATION_NONE;
```
ankuns: Bad type, use `nrf_802154_trx_transmit_notifications_t`

src/nrf_802154_core.c
```diff
@@ -741,7 +741,15 @@ static bool tx_init(const uint8_t * p_data, bool cca, bool disabled_was_triggere
...
+    if (nrf_802154_coex_tx_request_mode_get() == NRF_802154_COEX_TX_REQUEST_CCA_DONE)
+    {
+        notifications_mask |= TRX_TRANSMIT_NOTIFICATION_CCAIDLE;
+    }
+
```
ankuns: Could you mimic such behavior in `nrf_802154_core_transmit`:
- provide function `make_trx_frame_transmit_notification_mask` (simillar to `make_trx_frame_receive_notification_mask`)
- load result of the above funcction into `m_trx_transmit_frame_notifications_mask` in `nrf_802154_core_transmit`
- in `tx_init ` use saved `m_trx_transmit_frame_notifications_mask`

I know that at the moment we need to provide boot-time configuration only, but my proposal solves also run-time changing of the pib value. This solution is so simple. Also you can assert in `nrf_802154_trx_transmit_frame_ccaidle` that state is TX and `m_trx_transmit_frame_notifications_mask` contained TRX_TRANSMIT_NOTIFICATION_CCAIDLE

src/nrf_802154_trx.h
```diff
@@ -60,6 +60,15 @@ typedef enum
     TRX_RECEIVE_NOTIFICATION_STARTED = (1U << 1),
 } nrf_802154_trx_receive_notifications_t;
 
+typedef enum
```
ankuns: Please add comment
`/**@brief Notifications that can be enabled for @ref nrf_802154_trx_transmit_frame operation. */`

src/nrf_802154_core.c
```diff
@@ -741,7 +741,16 @@ static bool tx_init(const uint8_t * p_data, bool cca, bool disabled_was_triggere
         return false;
     }
 
-    nrf_802154_trx_transmit_frame(p_data, cca);
+    nrf_802154_trx_receive_notifications_t notifications_mask = TRX_TRANSMIT_NOTIFICATION_NONE;
+
+    if (nrf_802154_coex_tx_request_mode_get() == NRF_802154_COEX_TX_REQUEST_CCA_DONE)
```
jciupis: Should the core module call public driver API or PIB here?

src/nrf_802154_trx.h
```diff
@@ -278,7 +287,9 @@ bool nrf_802154_trx_receive_buffer_set(void * p_receive_buffer);
  *
  * @note To transmit ack after frame is received use @ref nrf_802154_trx_transmit_ack.
  */
-void nrf_802154_trx_transmit_frame(const void * p_transmit_buffer, bool cca);
+void nrf_802154_trx_transmit_frame(const void                            * p_transmit_buffer,
```
ankuns: Add doxygen `@param notifications_mask `

Update description (line 269) just after `- If cca succeded (channel was idle):` add that `nrf_802154_trx_transmit_frame_ccaidle` is called when cca==true and notifications_mask contained `TRX_TRANSMIT_NOTIFICATION_CCAIDLE`

src/nrf_802154_core.c
czeslawmakarski: Yes, using PIB would make more sense. Shall change.

src/nrf_802154_trx.h
```diff
@@ -262,23 +272,29 @@ bool nrf_802154_trx_receive_buffer_set(void * p_receive_buffer);
  *     - @ref nrf_802154_trx_transmit_frame_started handler is called from an ISR just after SHR is sent
  *       (only when @ref NRF_802154_TX_STARTED_NOTIFY_ENABLED == 1).
  *     - @ref nrf_802154_trx_transmit_frame_transmitted handler is called from an ISR after full frame is sent on air.
+ *     - If @ref TRX_TRANSMIT_NOTIFICATION_CCAIDLE was present in notifications_mask,
```
ankuns: Could it be moved to line 272. There is note about nrf_802154_trx_transmit_frame_started  and I the order matters here.

ankuns:
Regression test

PR #293, opened 2019-10-23
Pr/coex rx config core

ankuns:
Implemented coex RX requesting in core (modes ENERGY_DETECTION, PREAMBLE)

### DIFF
--- a/src/nrf_802154_core.c
+++ b/src/nrf_802154_core.c
@@ -72,26 +72,26 @@
 #include "nrf_802154_core_hooks.h"
 
 /// Delay before first check of received frame: 24 bits is PHY header and MAC Frame Control field.
-#define BCC_INIT           (3 * 8)
+#define BCC_INIT                 (3 * 8)
 
 /// Duration of single iteration of Energy Detection procedure
-#define ED_ITER_DURATION   128U
+#define ED_ITER_DURATION         128U
 /// Overhead of hardware preparation for ED procedure (aTurnaroundTime) [number of iterations]
-#define ED_ITERS_OVERHEAD  2U
+#define ED_ITERS_OVERHEAD        2U
 
-#define ACK_IFS            TURNAROUND_TIME ///< Ack Inter Frame Spacing [us] - delay between last symbol of received frame and first symbol of transmitted Ack
+#define ACK_IFS                  TURNAROUND_TIME ///< Ack Inter Frame Spacing [us] - delay between last symbol of received frame and first symbol of transmitted Ack
 
-#define MAX_CRIT_SECT_TIME 60              ///< Maximal time that the driver spends in single critical section.
+#define MAX_CRIT_SECT_TIME       60              ///< Maximal time that the driver spends in single critical section.
 
-#define LQI_VALUE_FACTOR   4               ///< Factor needed to calculate LQI value based on data from RADIO peripheral
-#define LQI_MAX            0xff            ///< Maximal LQI value
+#define LQI_VALUE_FACTOR         4               ///< Factor needed to calculate LQI value based on data from RADIO peripheral
+#define LQI_MAX                  0xff            ///< Maximal LQI value
 
 /** Get LQI of given received packet. If CRC is calculated by hardware LQI is included instead of CRC
  *  in the frame. Length is stored in byte with index 0; CRC is 2 last bytes.
  */
-#define RX_FRAME_LQI(data) ((data)[(data)[0] - 1])
+#define RX_FRAME_LQI(data)       ((data)[(data)[0] - 1])
 
-#define PRESTARTED_TIMER_TIMEOUT        (160U)
+#define PRESTARTED_TIMER_TIMEOUT (160U)
 
 #if NRF_802154_RX_BUFFERS > 1
 /// Pointer to currently used receive buffer.
@@ -1063,6 +1063,7 @@ void nrf_802154_trx_receive_ack_started(void)
 static void on_rx_prestarted_timeout(void * p_context)
 {
     bool in_crit_sect;
+
     (void)p_context;
 
     /* If we were in critical section this handler would not be called.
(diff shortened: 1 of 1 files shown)


src/nrf_802154_core.c
```diff
@@ -90,6 +91,8 @@
  */
 #define RX_FRAME_LQI(data) ((data)[(data)[0] - 1])
 
+#define PRESTARTED_TIMER_TIMEOUT        (160U)
```
czeslawmakarski: I would add a unit at the end of the define name.

src/nrf_802154_trx.c
```diff
@@ -1988,6 +1988,20 @@ void nrf_802154_radio_irq_handler(void)
     assert(result);
     (void)result;
 
+#if defined(NRF_RADIO_EVENT_HELPER1)
```
jciupis: Why execute this piece of code twice? Is it intentional or a copy paste mistake?

src/nrf_802154_core.c
```diff
@@ -1021,16 +1035,71 @@ void nrf_802154_trx_receive_ack_started(void)
...
+    bool in_crit_sect;
+
+    (void)p_context;
+
+    /* If we were in critical section this handler would not be called.
+     * If we were outside of critical section this handler could be called,
+     * and nrf_802154_critical_section_enter must succeed.
+     * Justification:
+     * - If higher priority interrupt preempts this handler before it takes critical section, that
+     * interrupt may enter/exit critical section, but when it returns to this this handler
```
jciupis: typo: double `this`

src/nrf_802154_core.c
```diff
@@ -1021,16 +1035,71 @@ void nrf_802154_trx_receive_ack_started(void)
...
+}
+
 void nrf_802154_trx_receive_frame_prestarted(void)
 {
     assert(m_state == RADIO_STATE_RX);
-    // TODO: Implement coex requesting behavior
+    assert((m_trx_receive_frame_notifications_mask & TRX_RECEIVE_NOTIFICATION_PRESTARTED) != 0U);
+
+    /* This handler serves one main purpose: boosting preconditions for receive.
+     * This handler may be not followed by nrf_802154_trx_receive_frame_started.
```
jciupis: comment clarification suggestion: `might not be followed`

src/nrf_802154_core.c
```diff
@@ -1062,6 +1131,7 @@ uint8_t nrf_802154_trx_receive_frame_bcmatched(uint8_t bcc)
             {
                 m_flags.frame_filtered = true;
 
+                /* Request boosted preceonditions */
```
jciupis: typo: `preconditions`

src/nrf_802154_core.c
```diff
@@ -1021,16 +1035,71 @@ void nrf_802154_trx_receive_ack_started(void)
...
 
 void nrf_802154_trx_receive_frame_started(void)
 {
     assert(m_state == RADIO_STATE_RX);
-    // TODO: Implement coex requesting behavior
+    assert((m_trx_receive_frame_notifications_mask & TRX_RECEIVE_NOTIFICATION_STARTED) != 0U);
+
+    nrf_802154_timer_sched_remove(&m_rx_prestarted_timer, NULL);
+
+    /* Request boosted preceonditions */
```
jciupis: typo: `preconditions`

src/nrf_802154_core.c
```diff
@@ -1021,16 +1035,71 @@ void nrf_802154_trx_receive_ack_started(void)
...
+
+    /* This handler serves one main purpose: boosting preconditions for receive.
+     * This handler may be not followed by nrf_802154_trx_receive_frame_started.
+     * That's why we need to revert boosted precondition if nrf_802154_trx_receive_frame_started
+     * doesn't come. We use timer for this purpose.
+     */
+
+    uint32_t now = nrf_802154_timer_sched_time_get();
+
+    nrf_802154_timer_sched_remove(&m_rx_prestarted_timer, NULL);
```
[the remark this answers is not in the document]
ankuns: It is crucial to remove timer FIRST, THEN modify `m_rx_prestarted_timer` struct, THEN add modified timer. If we didn't remove the timer we would modify the timer struct while the timer is armed. I think this would be an undefined behavior.

src/nrf_802154_core.c
```diff
@@ -625,6 +633,12 @@ static bool current_operation_terminate(nrf_802154_term_t term_lvl,
         {
             nrf_802154_trx_abort();
 
+            if (m_state == RADIO_STATE_RX)
+            {
+                nrf_802154_timer_sched_remove(&m_rx_prestarted_timer, NULL);
+                request_preconditions_for_state(m_state);
```
jciupis: This line is tricky, it looks like a mistake at the first glance and I think it deserves a comment. The comment should explain that it might happen that despite being in `RADIO_STATE_RX` which by default has `RSCH_PRIO_IDLE_LISTENING` preconditions, we might have `RSCH_PRIO_RX` preconditions here, so we're essentially dropping the boosted preconditions here.

src/nrf_802154_core.c
```diff
@@ -1021,16 +1035,71 @@ void nrf_802154_trx_receive_ack_started(void)
...
+     * This handler may be not followed by nrf_802154_trx_receive_frame_started.
+     * That's why we need to revert boosted precondition if nrf_802154_trx_receive_frame_started
+     * doesn't come. We use timer for this purpose.
+     */
+
+    uint32_t now = nrf_802154_timer_sched_time_get();
+
+    nrf_802154_timer_sched_remove(&m_rx_prestarted_timer, NULL);
+
+    /* Request boosted preceonditions */
```
jciupis: typo: `preconditions`

src/nrf_802154_core.c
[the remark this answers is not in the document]
hubertmis: I think this code is executed from the critical section and because of that the timer callback cannot cause a race here. But I'm ok to keep it as it is to make this function safer.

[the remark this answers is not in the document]
ankuns: It's safer as it is now.

src/nrf_802154_core.c
ankuns: done

src/nrf_802154_trx.c
ankuns: Very good catch, this modification leaked somehow during rebasing. Fixed now.

PR #294, opened 2019-10-23
Tied TX modes with relevant cases

czeslawmakarski:


src/nrf_802154_core.c
```diff
@@ -417,6 +417,8 @@ static bool ed_iter_setup(uint32_t * p_requested_ed_time_us, uint32_t * p_next_t
 
 static rsch_prio_t min_required_rsch_prio(radio_state_t state)
 {
+    nrf_802154_coex_tx_request_mode_t tx_req_mode = nrf_802154_pib_coex_tx_request_mode_get();
```
ankuns: Could it use `m_coex_tx_request_mode` which would be loaded from the pib in `nrf_802154_core_transmit` as it is done with `m_trx_transmit_frame_notifications_mask`? This way it would still allow to switch pib_coex_text_request_mode on the fly without altering current operation.

src/nrf_802154_core.c
```diff
@@ -1500,7 +1511,9 @@ void nrf_802154_trx_transmit_frame_ccaidle(void)
 {
     assert(m_state == RADIO_STATE_TX);
     assert(m_trx_transmit_frame_notifications_mask & TRX_TRANSMIT_NOTIFICATION_CCAIDLE);
-    // TODO: Implement coex requesting behavior
+    assert(nrf_802154_pib_coex_tx_request_mode_get() == NRF_802154_COEX_TX_REQUEST_CCA_DONE);
```
ankuns: Could it assert on `m_coex_tx_request_mode` which would be loaded from the pib in `nrf_802154_core_transmit` as it is done with `m_trx_transmit_frame_notifications_mask`?

src/nrf_802154_core.c
```diff
@@ -1672,6 +1684,7 @@ bool nrf_802154_core_transmit(nrf_802154_term_t              term_lvl,
         {
             state_set(cca ? RADIO_STATE_CCA_TX : RADIO_STATE_TX);
             mp_tx_data                              = p_data;
+            m_coex_tx_request_mode                  = nrf_802154_pib_coex_tx_request_mode_get();
```
ankuns: Please move this load (can be done together with `m_trx_transmit_frame_notifications_mask`). Before `state_set`. `state_set` calls `min_required_rsch_prio` which processes `m_coex_tx_request_mode` which has not been stored yet.

PR #295, opened 2019-10-23
Extend RSCH to support CSMA/CA backoffs

jciupis:
Exising implementation of delayed operations requests their preconditions as late as possible. This commit extends RSCH to support delayed timeslots with preconditions requested immediately and released with an explicit call only.

### DIFF
--- a/src/mac_features/nrf_802154_delayed_trx.c
+++ b/src/mac_features/nrf_802154_delayed_trx.c
@@ -201,17 +201,25 @@ static bool dly_op_request(uint32_t         t0,
                            uint32_t         length,
                            rsch_dly_ts_id_t dly_ts_id)
 {
-    bool        result;
-    rsch_prio_t prio;
+    bool                result;
+    rsch_dly_ts_param_t dly_ts_param =
+    {
+        .t0                = t0,
+        .dt                = dt,
+        .length            = length,
+        .prio              = RSCH_PRIO_IDLE,
+        .id                = dly_ts_id,
+        .prec_req_strategy = RSCH_PREC_REQ_STRATEGY_SHORTEST,
+    };
 
     switch (dly_ts_id)
     {
         case RSCH_DLY_TX:
-            prio = RSCH_PRIO_TX;
+            dly_ts_param.prio = RSCH_PRIO_TX;
             break;
 
         case RSCH_DLY_RX:
-            prio = RSCH_PRIO_RX;
+            dly_ts_param.prio = RSCH_PRIO_IDLE_LISTENING;
             break;
 
         default:
@@ -223,12 +231,7 @@ static bool dly_op_request(uint32_t         t0,
     // immediatly and interrupts current function execution.
     dly_op_state_set(dly_ts_id, DELAYED_TRX_OP_STATE_STOPPED, DELAYED_TRX_OP_STATE_PENDING);
 
-    result = nrf_802154_rsch_delayed_timeslot_request(t0,
-                                                      dt,
-                                                      length,
-                                                      prio,
-                                                      dly_ts_id,
-                                                      RSCH_PREC_REQ_MINIMAL);
+    result = nrf_802154_rsch_delayed_timeslot_request(&dly_ts_param);
 
     if (!result)
     {
@@ -490,7 +493,7 @@ void nrf_802154_rsch_delayed_timeslot_started(rsch_dly_ts_id_t dly_ts_id)
 
 bool nrf_802154_delayed_trx_transmit_cancel(void)
 {
-    bool result = nrf_802154_rsch_delayed_timeslot_cancel(RSCH_DLY_TX, RSCH_PREC_REQ_MINIMAL);
+    bool result = nrf_802154_rsch_delayed_timeslot_cancel(RSCH_DLY_TX);
 
     m_dly_op_state[RSCH_DLY_TX] = DELAYED_TRX_OP_STATE_STOPPED;
 
@@ -499,7 +502,7 @@ bool nrf_802154_delayed_trx_transmit_cancel(void)
 
 bool nrf_802154_delayed_trx_receive_cancel(void)
 {
-    bool result = nrf_802154_rsch_delayed_timeslot_cancel(RSCH_DLY_RX, RSCH_PREC_REQ_MINIMAL);
+    bool result = nrf_802154_rsch_delayed_timeslot_cancel(RSCH_DLY_RX);
     bool was_running;
 
     nrf_802154_timer_sched_remove(&m_timeout_timer, &was_running);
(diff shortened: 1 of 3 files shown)


src/rsch/nrf_802154_rsch.c
```diff
@@ -130,15 +134,18 @@ static rsch_prio_t max_prio_for_delayed_timeslot_get(void)
...
+        bool       active_dly_ts = p_dly_ts->prio > result;
+        uint32_t   t0            = p_dly_ts->t0;
+        uint32_t   dt            = p_dly_ts->dt - PREC_RAMP_UP_TIME -
+                                   nrf_802154_timer_sched_granularity_get();
 
-        if ((p_dly_ts->prio > result) && !nrf_802154_timer_sched_time_is_in_future(now, t0, dt))
+        if (p_dly_ts->prec_req == RSCH_PREC_REQ_MINIMAL)
         {
-            result = p_dly_ts->prio;
+            active_dly_ts = active_dly_ts && !nrf_802154_timer_sched_time_is_in_future(now, t0, dt);
```
hubertmis: These conditions are a little complicated now. How about something simpler like:
```
dly_ts_prio = p_dly_ts->prio;

if (p_dly_ts->pre_req == ...MINIMAL && !nrf_802154_timer_sched_time_is_in_future(now, t0, dt)) dly_ts_prio = IDLE;

result = result > dly_ts_prio ? result : dly_ts_prio'
```

src/mac_features/nrf_802154_delayed_trx.c
```diff
@@ -201,7 +201,23 @@ static bool dly_op_request(uint32_t         t0,
...
+    rsch_prio_t prio;
+
+    switch (dly_ts_id)
+    {
+        case RSCH_DLY_TX:
+            prio = RSCH_PRIO_TX;
+            break;
+
+        case RSCH_DLY_RX:
+            prio = RSCH_PRIO_RX;
```
hubertmis: I'm not sure if `RX` or `IDLE_LISTENING` here. I would select `IDLE_LISTENING`, but as we don't have requirements `RX` is also ok for me.

src/rsch/nrf_802154_rsch.h
```diff
@@ -90,10 +90,20 @@ typedef enum
...
     RSCH_DLY_TS_NUM, ///< Number of delayed timeslots.
 } rsch_dly_ts_id_t;
 
+/**
+ * @brief Enumeration of the precondition requesting strategies.
+ */
+typedef enum
+{
+    RSCH_PREC_REQ_CONSTANT, ///< Preconditions requested immediately and released only by @ref nrf_802154_rsch_delayed_timeslot_cancel call.
+    RSCH_PREC_REQ_MINIMAL,  ///< Preconditions requested as late as possible and released as soon as possible.
```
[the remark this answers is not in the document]
jciupis: That was my initial idea at first. However, I then tried to sneak an information not only about how the preconditions are requested (immediately vs last moment), but also about how they are released (held until explicit release vs as soon as possible) and that's how the current names came up. I see no problem with changing them, but I'd rather choose names that capture both sides of the process (i.e. request and release).

src/rsch/nrf_802154_rsch.c
```diff
@@ -487,7 +517,8 @@ bool nrf_802154_rsch_delayed_timeslot_request(uint32_t         t0,
     return result;
 }
 
-bool nrf_802154_rsch_delayed_timeslot_cancel(rsch_dly_ts_id_t dly_ts_id)
+bool nrf_802154_rsch_delayed_timeslot_cancel(rsch_dly_ts_id_t       dly_ts_id,
+                                             rsch_dly_ts_prec_req_t dly_ts_prec_req)
```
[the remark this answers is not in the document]
jciupis: That's correct, will fix that

src/mac_features/nrf_802154_delayed_trx.c
jciupis: I think `IDLE_LISTENING` is the better choice here

src/rsch/nrf_802154_rsch.c
```diff
@@ -424,11 +435,12 @@ bool nrf_802154_rsch_timeslot_request(uint32_t length_us)
...
-                                              uint32_t         dt,
-                                              uint32_t         length,
-                                              rsch_prio_t      prio,
-                                              rsch_dly_ts_id_t dly_ts_id)
+bool nrf_802154_rsch_delayed_timeslot_request(uint32_t               t0,
+                                              uint32_t               dt,
+                                              uint32_t               length,
+                                              rsch_prio_t            prio,
+                                              rsch_dly_ts_id_t       dly_ts_id,
+                                              rsch_dly_ts_prec_req_t dly_ts_prec_req)
```
ankuns: Please consider putting these growing argument list into a structure
`bool nrf_802154_rsch_delayed_timeslot_request(const nrf_802154_rsch_delayed_timeslot_request_t * p_req)`

src/rsch/nrf_802154_rsch.h
```diff
@@ -90,10 +90,20 @@ typedef enum
...
+} rsch_dly_ts_prec_req_t;
```
ankuns: Could this type be named `rsch_dly_ts_prec_req_strategy_t`, and enumerators have prefix `RSCH_PREC_REQ_STRATEGY_`.
Then in `nrf_802154_rsch_delayed_timeslot_request` parameter `dly_ts_prec_req` could be renamed into simple meaningful `prec_req_strategy`.

src/rsch/nrf_802154_rsch.h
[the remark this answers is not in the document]
hubertmis: I think it is hard to provide all information in these names. I can imagine only something like `RSCH_PREC_REQ_FROM_NOW_UNTIL_RELEASED` or `RSCH_PREC_REQ_FROM_LAST_POSSIBLE_MOMENT_UNTIL_OPERATION_STARTS` - I'm not sure if it is possible to fit it in a single line :)
I think it's better to provide one information in these names instead of enigmatic `_CONSTANT` and `_MINIMAL`.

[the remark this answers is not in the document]
hubertmis: I think `_SHORTEST` instead of `_MINIMAL` would be a better choice and `_STRATEGY_MANUAL` sounds good. Anyway, I'm ok with the naming. It's just my preference

src/rsch/nrf_802154_rsch.c
```diff
@@ -424,59 +436,73 @@ bool nrf_802154_rsch_timeslot_request(uint32_t length_us)
...
 
-bool nrf_802154_rsch_delayed_timeslot_request(uint32_t         t0,
-                                              uint32_t         dt,
-                                              uint32_t         length,
-                                              rsch_prio_t      prio,
-                                              rsch_dly_ts_id_t dly_ts_id)
+bool nrf_802154_rsch_delayed_timeslot_request(rsch_dly_ts_param_t * p_dly_ts_param)
 {
-    (void)length;
+    (void)p_dly_ts_param->length;
```
ankuns: Unnecessary access

PR #303, opened 2019-10-31
Pr/coex disable enable

ankuns:
Public api to enable/disable coex
Tweaks in core


src/nrf_802154.h
```diff
@@ -1241,6 +1241,37 @@ void nrf_802154_ack_timeout_set(uint32_t time);
...
  */
 
+/**
+ * @brief Enables wifi coex signaling.
+ *
+ * When @ref nrf_802154_init is called, the wifi coex signaling is initially disabled.
+ * You can call this function (after @ref nrf_802154_init) to enable the wifi coex signaling.
+ * When wifi coex signaling has been already enabled, this function has no effect.
+ *
+ * @note The nRF 802.15.4 driver must be in sleep state to perform this operations.
```
jciupis: What happens if this condition is not met? Is the driver left in an undefined state or the function simply has no effect? It's worth mentioning here.

src/nrf_802154.h
```diff
@@ -1241,6 +1241,37 @@ void nrf_802154_ack_timeout_set(uint32_t time);
...
+ * @retval false    Wifi coex is not supported.
+ */
+bool nrf_802154_wifi_coex_enable(void);
+
+/**
+ * @brief Disables wifi coex signaling.
+ *
+ * You can call this function (after @ref nrf_802154_init) to disable the wifi coex signaling.
+ * When wifi coex signaling has been already disabled, this function has no effect.
+ * @note The nRF 802.15.4 driver must be in sleep state to perform this operations.
```
jciupis: Missing newline before `@note`

src/nrf_802154.h
ankuns: doc updated

PR #306, opened 2019-11-04
Configurable CSMA-CA parameters

jciupis:


src/nrf_802154_pib.c
```diff
@@ -357,3 +382,54 @@ nrf_802154_coex_tx_request_mode_t nrf_802154_pib_coex_tx_request_mode_get(void)
 {
     return m_data.coex.tx_request_mode;
 }
+
+bool nrf_802154_pib_csmaca_min_be_set(uint8_t min_be)
+{
+    bool result = in_range(min_be, CSMACA_MIN_BE_MINIMUM, m_data.csmaca.max_be);
```
ankuns: I would remove current validation here, and I would leave validation for driver design limits only (min_be <= 8)

src/nrf_802154_pib.c
```diff
@@ -357,3 +382,54 @@ nrf_802154_coex_tx_request_mode_t nrf_802154_pib_coex_tx_request_mode_get(void)
...
+}
+
+uint8_t nrf_802154_pib_csmaca_min_be_get(void)
+{
+    return m_data.csmaca.min_be;
+}
+
+bool nrf_802154_pib_csmaca_max_be_set(uint8_t max_be)
+{
+    bool result = in_range(max_be, CSMACA_MAX_BE_MINIMUM, CSMACA_MAX_BE_MAXIMUM);
```
ankuns: I would remove current validation here, and I would leave validation for driver design limits only (max_be <= 8)

src/nrf_802154_pib.c
```diff
@@ -46,6 +46,14 @@
...
 #include "fal/nrf_802154_fal.h"
 
+#define CSMACA_MIN_BE_MINIMUM       0 ///< The minimum value of the backoff exponent (BE) allowed by the protocol specification
+
+#define CSMACA_MAX_BE_MINIMUM       3 ///< The minimum value of the maximum value of the backoff exponent (BE) allowed by the protocol specification
+#define CSMACA_MAX_BE_MAXIMUM       8 ///< The maximum value of the backoff exponent (BE) allowed by the protocol specification
+
+#define CSMACA_MAX_BACKOFFS_MINIMUM 0 ///< The minimum value of the maximum number of CSMA-CA backoffs allowed by the protocol specification
+#define CSMACA_MAX_BACKOFFS_MAXIMUM 5 ///< The maximum number of the CSMA-CA backoffs allowed by the protocol specification
+
```
ankuns: These limits will disallow to experiment with values out of current protocol, what might be objective of configurable parameters feature.

test/unit tests/csma_ca/unity_test.c
```diff
@@ -0,0 +1,318 @@
...
+    }
+}
+
+static void verify_channel_empty_scenario(uint8_t min_be, uint8_t max_be)
+{
+    for (int i = 0; i < COEX_TX_REQUEST_MODE_NUM; i++)
+    {
+        for (uint8_t max_backoffs = CSMACA_MAX_BACKOFFS_MINIMUM; max_backoffs <= CSMACA_MAX_BACKOFFS_MAXIMUM; max_backoffs++)
+        {
+            uint8_t failed_backoffs = 0;
```
ankuns: Consider `cca_busy_backoffs` name. The word `failed` in context of unit tests suggests some failed test.

src/nrf_802154_pib.h
```diff
@@ -220,6 +220,59 @@ bool nrf_802154_pib_coex_tx_request_mode_set(nrf_802154_coex_tx_request_mode_t m
  */
 nrf_802154_coex_tx_request_mode_t nrf_802154_pib_coex_tx_request_mode_get(void);
 
+/**
+ * @brief Sets the minimum value of the backoff exponent (BE) in the CSMA-CA algorithm.
+ *
+ * @param[in] min_be  Minimum value of the backoff exponent.
+ *
+ * @retval true    When value provided by @p min_be is allowed by the protocol.
```
ankuns: " is allowed by the protocol." this would require change if you decide to accept values out of protocol.

src/nrf_802154_config.h
```diff
@@ -293,36 +293,36 @@ extern "C" {
 #endif
 
 /**
- * @def NRF_802154_CSMA_CA_MIN_BE
+ * @def NRF_802154_CSMA_CA_MIN_BE_DEFAULT
```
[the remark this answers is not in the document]
LuDuda: In OpenThread whenever we change the define name we always add a check ensuring the old define is no longer used. Maybe you can adopt this approach. Please look at: https://github.com/openthread/openthread/blob/master/src/core/config/openthread-core-config-check.h

src/mac_features/nrf_802154_csma_ca.c
```diff
@@ -145,6 +144,14 @@ static void random_backoff_start(void)
...
     uint8_t backoff_periods = nrf_802154_random_get() % (1 << m_be);
 
+    // If maximum number of CSMA-CA backoffs is equal to 0, this function is called only once
+    // and no more backoffs will follow. Forcing the first and only backoff to 0 has the same
+    // effect as no backoff at all.
+    if (0 == nrf_802154_pib_csmaca_max_backoffs_get())
+    {
+        backoff_periods = 0;
+    }
+
```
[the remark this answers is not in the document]
jciupis: The current approach is more composed in terms of number of lines, the values are zeroes ( = no magic numbers) and there's a comment explaining why the variable needs to be pulled down to 0 in a specific case. I'd prefer to leave it as it is, but if you have a strong opinion on that, I'll change it

jciupis:
@LuDuda It hasn't been tested yet, but I see no problem in changing CSMA/CA configuration in PIB for every transmitted frame, **provided that no CSMA/CA transactions are running at the moment**. If the CSMA/CA module is idle and has no activity scheduled, I reckon that the configuration can be modified safely.

src/nrf_802154_config.h
```diff
@@ -293,36 +293,55 @@ extern "C" {
...
  *
- * The minimum value of the backoff exponent (BE) in the CSMA-CA algorithm
+ * The default minimum value of the backoff exponent (BE) in the CSMA-CA algorithm
  * (see IEEE 802.15.4-2015: 6.2.5.1).
  *
+ * @note The value of this parameter may be changed from default by calling
```
ankuns: Rather `The value of backoff exponent (BE) may be changed from default by calling`. This `NRF_802154_CSMA_CA_MIN_BE` parameter cannot be changed at runtime as it is compile-time.

ankuns: There are also such mental abbreviations in other places.

src/nrf_802154_pib.c
```diff
@@ -357,3 +369,33 @@ nrf_802154_coex_tx_request_mode_t nrf_802154_pib_coex_tx_request_mode_get(void)
 {
     return m_data.coex.tx_request_mode;
 }
+
+void nrf_802154_pib_csmaca_min_be_set(uint8_t min_be)
+{
+    m_data.csmaca.min_be = min_be;
```
ankuns: Please just validate for min_be <= 8, as this is an implementation limit of the driver

src/nrf_802154_pib.c
```diff
@@ -357,3 +369,33 @@ nrf_802154_coex_tx_request_mode_t nrf_802154_pib_coex_tx_request_mode_get(void)
...
+}
+
+uint8_t nrf_802154_pib_csmaca_min_be_get(void)
+{
+    return m_data.csmaca.min_be;
+}
+
+void nrf_802154_pib_csmaca_max_be_set(uint8_t max_be)
+{
+    m_data.csmaca.max_be = max_be;
```
ankuns: Please just validate for max_be <= 8, as this is an implementation limit of the driver

PR #307, opened 2019-11-06
Interframe space

czeslawmakarski:
Implemented Interframe space functionality.


src/nrf_802154_pib.h
```diff
@@ -270,6 +270,48 @@ void nrf_802154_pib_csmaca_max_backoffs_set(uint8_t max_backoffs);
...
 
+/**
+ * @brief Gets IFS trigger only on address match value.
+ *
+ * @return Current IFS trigger on address match value.
+ */
+bool nrf_802154_pib_ifs_address_match_only_get(void);
+
+/**
+ * @brief Sets IFS trigger only on address match value.
```
jciupis: The documentation is unclear about whether `only on address match` means that otherwise IFS is used always or never. It's crucial information and should be clarified in the documentation.

src/mac_features/nrf_802154_ifs.c
```diff
@@ -0,0 +1,212 @@
...
+    {
+        return true;
+    }
+
+    if (is_extended == m_is_last_address_extended)
+    {
+        uint8_t * last_addr = is_extended ? m_last_extended_address : m_last_short_address;
+        size_t addr_len     = is_extended ? EXTENDED_ADDRESS_SIZE : SHORT_ADDRESS_SIZE;
+
+        if (!memcmp(addr, last_addr, addr_len))
```
jciupis: Please change this condition to `0 == memcmp`. The amount of logic in the module is enough to melt one's brain anyway and IMO it's easier to follow if the "memory equal" return value of memcmp is used explicitly here instead of a negation of an integer

src/mac_features/nrf_802154_ifs.c
```diff
@@ -0,0 +1,212 @@
...
+    if (nrf_802154_pib_ifs_address_match_only_get() == false)
+    {
+        /* We dont't care about the address match. */
+        return true;
+    }
+
+    bool is_extended;
+    const uint8_t * addr = nrf_802154_frame_parser_dst_addr_get(p_frame, &is_extended);
+    if (!addr)
+    {
```
jciupis: Please add a comment describing why we need IFS when address could not be retrieved, similar to the one in line 79

src/mac_features/nrf_802154_ifs.c
```diff
@@ -0,0 +1,212 @@
...
+        }
+    }
+
+    return false;
+}
+
+/**@brief Checks if the IFS is needed by measuring time between the actual and the last frames. */
+static bool is_ifs_needed_by_time(const uint8_t * p_frame, bool * is_lifs)
+{
+    assert(m_current_timestamp > m_last_frame_timestamp);
```
jciupis: What if the timestamp wraps around?

src/mac_features/nrf_802154_ifs.c
```diff
@@ -0,0 +1,212 @@
...
+
+static void callback_fired(void * p_context)
+{
+    ifs_operation_t * context = (ifs_operation_t *)p_context;
+    nrf_802154_request_transmit(NRF_802154_TERM_802154,
+                                REQ_ORIG_IFS,
+                                context->p_data,
+                                context->cca,
+                                true,
+                                NULL);
```
jciupis: Please correct me if I'm wrong, but `immediate=true` and `notify_function=NULL` seems like a bad combination to me. If transmission attempt fails for whatever reason, the frame is going to be dropped silently. Is that fine? It might be, I'm not sure

src/mac_features/nrf_802154_ifs.c
```diff
@@ -0,0 +1,212 @@
...
+        memcpy(m_last_short_address, addr, SHORT_ADDRESS_SIZE);
+    }
+
+    m_last_frame_length = p_frame[0];
+    assert(m_last_frame_length < 128);
+}
+
+bool nrf_802154_ifs_abort_hook(nrf_802154_term_t term_lvl, req_originator_t req_orig)
+{
+    bool result = true;
```
jciupis: Missing newline

src/mac_features/nrf_802154_ifs.c
```diff
@@ -0,0 +1,212 @@
...
+    if (req_orig == REQ_ORIG_IFS)
+    {
+        // Ignore if self-request.
+    }
+    else if (nrf_802154_timer_sched_is_running(&m_timer))
+    {
+        if (term_lvl >= NRF_802154_TERM_802154)
+        {
+            ifs_operation_t * op = (ifs_operation_t *)m_timer.p_context;
```
jciupis: Pointers should be prefixed with `p_`

src/mac_features/nrf_802154_ifs.c
```diff
@@ -0,0 +1,212 @@
...
+            bool was_running;
+            nrf_802154_notify_transmit_failed(op->p_data, NRF_802154_TX_ERROR_ABORTED);
+            nrf_802154_timer_sched_remove(&m_timer, &was_running);
```
jciupis: Is this order of calls intentional? It would seem more natural for the IFS module to do its job first (stop the timer) and pass the control to the higher layer later (notify about failure). What do you think?

src/nrf_802154_const.h
```diff
@@ -158,6 +158,10 @@
 
 #define BROADCAST_ADDRESS            ((uint8_t[SHORT_ADDRESS_SIZE]) {0xff, 0xff}) ///< Broadcast short address.
 
+#define MIN_SIFS_PERIOD_US           192                                          ///< Minimum Short IFS period default value in us.
+#define MIN_LIFS_PERIOD_US           640                                          ///< Minimum Short IFS period default value in us.
```
jciupis: `Short` instead of `Long`

src/nrf_802154_core.c
```diff
@@ -1820,25 +1820,28 @@ bool nrf_802154_core_transmit(nrf_802154_term_t              term_lvl,
...
+                m_trx_transmit_frame_notifications_mask = make_trx_frame_transmit_notification_mask();
 
-            state_set(cca ? RADIO_STATE_CCA_TX : RADIO_STATE_TX);
-            mp_tx_data = p_data;
+                state_set(cca ? RADIO_STATE_CCA_TX : RADIO_STATE_TX);
+                mp_tx_data = p_data;
 
-            if (immediate)
-            {
                 result = tx_init(p_data, cca, true);
```
jciupis: I think it makes more sense to have this line under the if clause as it was before.

src/mac_features/nrf_802154_ifs.h
```diff
@@ -0,0 +1,71 @@
...
+ * @brief Aborts an ongoing IFS-delayed transmission.
+ *
+ * @param[in]  term_lvl  Termination level set by the request to abort the ongoing operation.
+ * @param[in]  req_orig  Module that originates this request.
+ *
+ * @retval  true   Transmission procedure have been stopped.
+ * @retval  false  Transmission procedure were not running.
+ *
+ */
+bool nrf_802154_ifs_abort_hook(nrf_802154_term_t term_lvl, req_originator_t req_orig);
```
jciupis: I think `hook` can be removed from this function name safely to align with other aborts (csma_ca, delayed_trx etc.)

src/nrf_802154_pib.c
```diff
@@ -63,21 +63,33 @@ typedef struct
...
-    uint8_t                 short_addr[SHORT_ADDRESS_SIZE];       ///< Short Address of this node.
-    uint8_t                 extended_addr[EXTENDED_ADDRESS_SIZE]; ///< Extended Address of this node.
-    nrf_802154_cca_cfg_t    cca;                                  ///< CCA mode and thresholds.
-    bool                    promiscuous : 1;                      ///< Indicating if radio is in promiscuous mode.
-    bool                    auto_ack    : 1;                      ///< Indicating if auto ACK procedure is enabled.
-    bool                    pan_coord   : 1;                      ///< Indicating if radio is configured as the PAN coordinator.
-    uint8_t                 channel     : 5;                      ///< Channel on which the node receives messages.
-    nrf_802154_pib_coex_t   coex;                                 ///< Coex-related fields.
+    bool     matching_addresses_only; //< Whether to use it for matching addresses only
+    uint16_t min_sifs_period_us;      //< 12 * 16 us = 192 us
```
jciupis: If there are getters and setters for these values, these comments are not very accurate

src/nrf_802154_pib.c
```diff
@@ -63,21 +63,33 @@ typedef struct
...
+    nrf_802154_pib_coex_t   coex;                                   ///< Coex-related fields.
     nrf_802154_pib_csmaca_t csmaca;                               ///< CSMA-CA related fields.
+    nrf_802154_pib_ifs_t    ifs;                                    ///< IFS-related fields.
 } nrf_802154_pib_data_t;
 
 // Static variables.
-static nrf_802154_pib_data_t m_data; ///< Buffer containing PIB data.
+static nrf_802154_pib_data_t m_data =
+{
+    .ifs.min_sifs_period_us = MIN_SIFS_PERIOD_US,
```
jciupis: Please move initialization to `nrf_802154_pib_init()`

src/mac_features/nrf_802154_ifs.c
```diff
@@ -0,0 +1,212 @@
...
+#include "mac_features/nrf_802154_frame_parser.h"
+#include "timer_scheduler/nrf_802154_timer_sched.h"
+
+typedef struct
+{
+    uint8_t * p_data;
+    bool cca;
+} ifs_operation_t;
+
+static uint32_t m_current_timestamp;
```
ankuns: `m_current_timestamp` global is unnecessary, you could pass timestamp as a function parameter where needed. See my other comment

src/mac_features/nrf_802154_ifs.c
```diff
@@ -0,0 +1,212 @@
...
+static uint8_t m_last_short_address[SHORT_ADDRESS_SIZE];
+static uint8_t m_last_extended_address[EXTENDED_ADDRESS_SIZE];
```
ankuns: `m_last_short_address` and `m_last_extended_address` could be in union to save ram

src/mac_features/nrf_802154_ifs.c
```diff
@@ -0,0 +1,212 @@
...
+                                context->p_data,
+                                context->cca,
+                                true,
+                                NULL);
+}
+
+/**@brief Checks if the IFS is needed by comparing the addresses of the actual and the last frames. */
+static bool is_ifs_needed_by_address(const uint8_t * p_frame)
+{
+    if (nrf_802154_pib_ifs_address_match_only_get() == false)
```
ankuns: Never explicit compare with `true`/`false`. Here `!nrf_802154_pib_ifs_address_match_only_get()` would be justified.

src/nrf_802154_config.h
```diff
@@ -407,6 +407,16 @@ extern "C" {
 #define NRF_802154_MAX_ACK_IE_SIZE 8
 #endif
 
+/**
+ * @def NRF_802154_IFS_ENABLED
+ *
+ * Indicates whether the Short/Long IFS feature is to be enabled in the driver.
```
ankuns: I would expand IFS into `Inter-frame spacing`

src/nrf_802154_config.h
```diff
@@ -407,6 +407,16 @@ extern "C" {
 #define NRF_802154_MAX_ACK_IE_SIZE 8
 #endif
 
+/**
```
ankuns: This is within `@defgroup nrf_802154_config_timeout` is it the case?

src/nrf_802154_core.c
```diff
@@ -1820,25 +1820,28 @@ bool nrf_802154_core_transmit(nrf_802154_term_t              term_lvl,
 
         if (result)
         {
-            m_coex_tx_request_mode                  = nrf_802154_pib_coex_tx_request_mode_get();
-            m_trx_transmit_frame_notifications_mask = make_trx_frame_transmit_notification_mask();
+            /* Short-circuit evaluation in place. */
+            if ((immediate) || (nrf_802154_core_hooks_pre_transmission(p_data, cca)))
```
ankuns: Should really `nrf_802154_core_hooks_pre_transmission`  be not called when `immediate==true` ? How one can extend those hooks?

src/mac_features/nrf_802154_ifs.c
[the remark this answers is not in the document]
czeslawmakarski: It's more uglify than prettify these days, but overall I agree.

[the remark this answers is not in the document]
jciupis: 3ffi0o

src/mac_features/nrf_802154_ifs.c
czeslawmakarski: Makes sense.

src/mac_features/nrf_802154_ifs.c
```diff
@@ -0,0 +1,212 @@
...
+    {
+        return true;
+    }
+
+    uint32_t dt = is_lifs ? nrf_802154_pib_ifs_min_lifs_period_get() : nrf_802154_pib_ifs_min_sifs_period_get();
+
+    m_context.p_data = (uint8_t *)p_frame;
+    m_context.cca = cca;
+    m_timer.t0 = m_current_timestamp;
+    m_timer.dt = dt;
```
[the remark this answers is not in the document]
czeslawmakarski: Not only ramp-up time, also the time of software to turn around a do the job is even more than 40 us. Hence, the decision is to treat IFS period as the 'at least' amount of time.

src/nrf_802154_pib.c
```diff
@@ -63,21 +63,33 @@ typedef struct
...
+    .ifs.min_lifs_period_us = MIN_LIFS_PERIOD_US
+}; ///< Buffer containing PIB data.
```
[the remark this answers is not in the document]
czeslawmakarski: Sure thing.

src/nrf_802154_core.c
czeslawmakarski: The idea is (at least that's how I understand that) that the `immediate=true` calls are the 'shoot straight away' and also 'clear to send'. This way no `pre_transmission` hooks are needed to be applied. Do you agree?

src/nrf_802154_pib.c
```diff
@@ -63,21 +63,33 @@ typedef struct
...
-    uint8_t                 pan_id[PAN_ID_SIZE];                  ///< Pan Id of this node.
-    uint8_t                 short_addr[SHORT_ADDRESS_SIZE];       ///< Short Address of this node.
-    uint8_t                 extended_addr[EXTENDED_ADDRESS_SIZE]; ///< Extended Address of this node.
-    nrf_802154_cca_cfg_t    cca;                                  ///< CCA mode and thresholds.
-    bool                    promiscuous : 1;                      ///< Indicating if radio is in promiscuous mode.
-    bool                    auto_ack    : 1;                      ///< Indicating if auto ACK procedure is enabled.
-    bool                    pan_coord   : 1;                      ///< Indicating if radio is configured as the PAN coordinator.
-    uint8_t                 channel     : 5;                      ///< Channel on which the node receives messages.
-    nrf_802154_pib_coex_t   coex;                                 ///< Coex-related fields.
+    bool     matching_addresses_only; //< Whether to use it for matching addresses only
```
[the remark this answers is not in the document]
czeslawmakarski: I wholeheartedly agree with you.

src/nrf_802154_pib.c
czeslawmakarski: Your wish is my command.

src/nrf_802154_pib.h
```diff
@@ -270,6 +270,48 @@ void nrf_802154_pib_csmaca_max_backoffs_set(uint8_t max_backoffs);
...
+ *
+ * @param[in] enabled IFS trigger on address match value.
+ */
+void nrf_802154_pib_ifs_address_match_only_set(bool enabled);
```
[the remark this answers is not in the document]
czeslawmakarski: A very useful idea indeed.

src/nrf_802154_pib.h
czeslawmakarski: This part shall be reworked nonetheless, but thank you for the notification.

src/nrf_802154_pib.c
czeslawmakarski: Indeed they are not.

src/nrf_802154_const.h
czeslawmakarski: Great catch!

src/mac_features/nrf_802154_ifs.h
czeslawmakarski: Absolutely - consistency is our primary goal.

src/mac_features/nrf_802154_ifs.h
```diff
@@ -0,0 +1,71 @@
...
+#include <stdbool.h>
+#include <stdint.h>
+
+#include "nrf_802154_const.h"
+#include "nrf_802154_types.h"
+
+/**
+ * @brief Catches the frame before transmission and checks if it needs to be delayed.
+ *
+ * @param[in]  p_frame  Pointer to the buffer that contains the transmitted frame.
```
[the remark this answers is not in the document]
czeslawmakarski: Precision is important. Thanks.

src/mac_features/nrf_802154_ifs.h
```diff
@@ -0,0 +1,71 @@
...
+#define NRF_802154_IFS_H
+
+#include <stdbool.h>
+#include <stdint.h>
+
+#include "nrf_802154_const.h"
+#include "nrf_802154_types.h"
+
+/**
+ * @brief Catches the frame before transmission and checks if it needs to be delayed.
```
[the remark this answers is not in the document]
czeslawmakarski: Thanks for `catch`ing this.

src/mac_features/nrf_802154_ifs.c
czeslawmakarski: Yes.

src/mac_features/nrf_802154_ifs.c
```diff
@@ -0,0 +1,212 @@
...
+    {
+        return false;
+    }
+
+    return true;
+}
+
+bool nrf_802154_ifs_pretransmission_hook(const uint8_t * p_frame, bool cca)
+{
+    m_current_timestamp = nrf_802154_timer_sched_time_get();
```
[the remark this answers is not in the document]
czeslawmakarski: It is evil indeed and we try to stand on the side of *good*. Worry no more, it shall be fixed.

src/mac_features/nrf_802154_ifs.c
czeslawmakarski: A special function shall be used to help us escape this predicament. Thanks for noticing!

src/mac_features/nrf_802154_ifs.c
```diff
@@ -0,0 +1,212 @@
...
+            return true;
+        }
+    }
+
+    return false;
+}
+
+/**@brief Checks if the IFS is needed by measuring time between the actual and the last frames. */
+static bool is_ifs_needed_by_time(const uint8_t * p_frame, bool * is_lifs)
+{
```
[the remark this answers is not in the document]
czeslawmakarski: It could and it will. I hope you shall approve of incoming changes.

src/mac_features/nrf_802154_ifs.c
czeslawmakarski: Say no more!

src/mac_features/nrf_802154_ifs.c
czeslawmakarski: Done!

src/mac_features/nrf_802154_ifs.c
czeslawmakarski: As discussed F2F: integer promotion can be tricky, so we prefer to be on the safe side. Thanks!

src/mac_features/nrf_802154_ifs.c
```diff
@@ -0,0 +1,212 @@
...
+static uint8_t m_last_extended_address[EXTENDED_ADDRESS_SIZE];
+static bool m_is_last_address_extended;
+static uint32_t m_last_frame_timestamp;
+static uint8_t m_last_frame_length;
+static ifs_operation_t m_context;
+static nrf_802154_timer_t m_timer;
+
+static void callback_fired(void * p_context)
+{
+    ifs_operation_t * context = (ifs_operation_t *)p_context;
```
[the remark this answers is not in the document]
czeslawmakarski: They should and they will.

src/mac_features/nrf_802154_ifs.c
czeslawmakarski: I'll add the callback, it won't hurt.

src/mac_features/nrf_802154_ifs.c
czeslawmakarski: Since we don't need to acccess them simultanteously, they indeed can be in the union. Thanks for helping to reduce memory footprint.

src/mac_features/nrf_802154_ifs.c
czeslawmakarski: I've seen your other comment and did react on it. Thanks.

src/mac_features/nrf_802154_ifs.c
```diff
@@ -0,0 +1,212 @@
...
+ *
+ */
+
+#include "nrf_802154_ifs.h"
+
+#include <assert.h>
+#include <stdint.h>
+#include <string.h>
+
+#include "nrf_802154_request.h"
```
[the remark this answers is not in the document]
czeslawmakarski: I doublechecked with `strcmp`, you are right. Shall be fixed.

src/mac_features/nrf_802154_ifs.c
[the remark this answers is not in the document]
czeslawmakarski: This snipped actually fits much better. I'll allow myself to copy it in the code. Thanks.

src/nrf_802154_config.h
czeslawmakarski: Not really - but thanks to your observation it shall be placed in the appropriate `@defgroup`. Much obliged.

src/nrf_802154_config.h
czeslawmakarski: You are right - AFAIK unnecessary acronyms are PITA.

src/nrf_802154_core.c
czeslawmakarski: Could you possibly explain a bit more? Whole rework of this core routine was made to match calls 1-to-1 (before and after), but maybe I don't understand something? We can discuss F2F.

src/nrf_802154_core.c
jciupis: It does not necessarily mean `clear to send`, but if it's not clear to send then the core is free to drop the frame. So I agree with @czeslawmakarski

src/nrf_802154_core.c
jciupis: My mistake, it's fine

src/mac_features/nrf_802154_ifs.c
```diff
@@ -101,60 +107,66 @@ static bool is_ifs_needed_by_address(const uint8_t * p_frame)
...
-static bool is_ifs_needed_by_time(const uint8_t * p_frame, bool * is_lifs)
+/**@brief Checks if the IFS is needed by measuring time between the actual and the last frames. 
+ *        Returns the needed ifs, 0 if none.
+ */
+static uint16_t ifs_needed_by_time(const uint8_t * p_frame, uint32_t current_timestamp)
 {
-    assert(m_current_timestamp > m_last_frame_timestamp);
-    uint32_t dt = m_current_timestamp - m_last_frame_timestamp;
-
+    assert(nrf_802154_timer_sched_time_is_in_future(m_last_frame_timestamp, 0, current_timestamp));
```
ankuns: I understand you idea here, but this assert can be detrimental.
Consider the case, when last frame has been transmitted so long ago (more than 2^31 us), that `nrf_802154_timer_sched_time_is_in_future` returns false due to integer overflow. Then the code asserts here, but it should't.

Proposed solutions:
- Simple. Just remove the assert. It is better to add one unwanted ifs delay if device transmits very rarely and go on than hang due to assert.
- Design change. Start IFS timer when previous frame has been transmitted. If the timer fires and no new transmit request has been requested during its period, the timer does nothing. If there is a new transmit request when timer is counting, check matching adresses. If no match, kill timer and transmit, If match, just save some static data to make timer's callback transmit when it fires.

@jciupis @hubertmis Please comment your opinion

src/nrf_802154_pib.c
```diff
@@ -224,6 +232,9 @@ void nrf_802154_pib_init(void)
     m_data.csmaca.min_be       = NRF_802154_CSMA_CA_MIN_BE_DEFAULT;
     m_data.csmaca.max_be       = NRF_802154_CSMA_CA_MAX_BE_DEFAULT;
     m_data.csmaca.max_backoffs = NRF_802154_CSMA_CA_MAX_CSMA_BACKOFFS_DEFAULT;
+
+    m_data.ifs.min_sifs_period_us = MIN_SIFS_PERIOD_US;
```
jciupis: It'd be nice to set the mode explicitly here as well.

src/nrf_802154_pib.h
```diff
@@ -271,18 +271,18 @@ void nrf_802154_pib_csmaca_max_backoffs_set(uint8_t max_backoffs);
...
 
 /**
- * @brief Sets IFS trigger only on address match value.
+ * @brief Sets IFS operation mode.
  *
- * @param[in] enabled IFS trigger on address match value.
+ * @param[in] mode IFS operation mode. Refer to @ref nrf_802154_ifs_mode_t for details.
  */
-void nrf_802154_pib_ifs_address_match_only_set(bool enabled);
+void nrf_802154_pib_ifs_mode_set(nrf_802154_ifs_mode_t mode);
```
ankuns: It would be great to validate for allowed values, and return `true` if success, and `false` if unsupported value is given (and no write to pib).

src/mac_features/nrf_802154_ifs.c
```diff
@@ -0,0 +1,229 @@
...
+    {
+        // No frame was transmitted before - skip the routine.
+        return true;
+    }
+
+    if (mode == NRF_802154_IFS_MODE_MATCHING_ADDRESSES && !is_ifs_needed_by_address(p_frame))
+    {
+        return true;
+    }
+
```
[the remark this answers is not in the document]
czeslawmakarski: You're right. Since this is a pretransmission hook, we want to timestamp as lately as possible.

src/nrf_802154_pib.h
czeslawmakarski: Makes total sense.

src/mac_features/nrf_802154_ifs.c
jciupis: I'm in favour of dropping the assert completely - as @ankuns said, it's not a big deal to send a frame with several hundred microseconds delay if the previous one was sent tens of minutes before that (and that's how long the timer scheduler takes to overflow, give or take).

czeslawmakarski: I agree that the problem exists. Its probability of occurence is quite low (maybe only Green Power device would send the frames rarely), but nevertheless it is a problem due to the bad design.

Now, about the resolutions.
Removing assert is trivial, naturally.
Your proposed design change would do the job, however it would make implementation less straightforward for the sake of complexity. Also I would suspect an increased power consumption (every transmission means one more callback to kick, so more CPU clicking and so on).

Let's wait for the comments.

hubertmis: I agree we cannot assert it here due to reason pointed by @ankuns .
How about following code?
```suggestion
    if (!nrf_802154_timer_sched_time_is_in_future(m_last_frame_timestamp, 0, current_timestamp))
    {
        return 0;
    }
```

czeslawmakarski: Hubert's resolution is fine by me. Everybody agrees?

src/mac_features/nrf_802154_ifs.c
```diff
@@ -0,0 +1,229 @@
...
+static uint32_t m_last_frame_timestamp;                                 ///< Timestamp of the last transmitted frame (end of frame).
+static uint8_t m_last_frame_length;                                     ///< Length in bytes of the last transmitted frame.
+static ifs_operation_t m_context;                                       ///< Context passed to the timer.
+static nrf_802154_timer_t m_timer;                                      ///< Interframe space timer.
+
+static void ifs_tx_result_notify(bool result)
+{
+    if (!result)
+    {
+        nrf_802154_notify_transmit_failed(m_context.p_data, NRF_802154_TX_ERROR_BUSY_CHANNEL);
```
jciupis: I think `BUSY_CHANNEL` is rather inaccurate here. In my opinion either `ABORTED` or `TIMESLOT_DENIED` should be the error code notified here.

czeslawmakarski: In general I agree, however do we have a clear understanding why the immediate transmission did (or could) fail?

src/mac_features/nrf_802154_ifs.c
jciupis: *(in Aragorn's voice)* You have my sword!

src/mac_features/nrf_802154_ifs.c
jciupis: We have a clear understanding that it doesn't fail due to a busy channel ;) but no, currently there's no way to find out the exact reason

czeslawmakarski: Okay, I'll make it a `TIMESLOT_DENIED`.

src/nrf_802154_pib.c
ankuns: ...under `#if`, to save flash, 
```c
#if (NRF_802154_IFS_MODE_DISABLED != 0U)
m_data.ifs.mode = NRF_802154_IFS_MODE_DISABLED;
#endif
```

src/mac_features/nrf_802154_ifs.c
ankuns: ok, go with code proposed by @hubertmis

src/nrf_802154_pib.c
czeslawmakarski: @ankuns Did you by any chance mean:
```
#if NRF_802154_IFS_ENABLED
m_data.ifs.mode = NRF_802154_IFS_MODE_DISABLED;
#endif
```
?

src/nrf_802154_core_hooks.c
```diff
@@ -151,6 +167,28 @@ bool nrf_802154_core_hooks_terminate(nrf_802154_term_t term_lvl, req_originator_
     return result;
 }
 
+bool nrf_802154_core_hooks_pre_transmission(const uint8_t * p_frame, bool cca)
+{
+    bool result = true;
+
+    for (uint32_t i = 0; i < sizeof(m_pre_transmission_hooks) / sizeof(m_pre_transmission_hooks[0]); i++)
+    {
+        if (m_pre_transmission_hooks[i] == NULL)
```
ankuns: If the loop goes over existing items `i < sizeof(m_pre_transmission_hooks) / sizeof(m_pre_transmission_hooks[0])`, this check'n'break is not necessary. If you decide to remove, please remove also `NULL` from the `m_pre_transmission_hooks`.

src/mac_features/nrf_802154_ifs.c
```diff
@@ -0,0 +1,229 @@
...
+    {
+        memcpy(m_last_address.ext, addr, EXTENDED_ADDRESS_SIZE);
+    }
+    else
+    {
+        memcpy(m_last_address.sh, addr, SHORT_ADDRESS_SIZE);
+    }
+
+    m_last_frame_length = p_frame[0];
+    assert(m_last_frame_length < 128);
```
ankuns: This assert is also unnecessary. Nowhere in the code you provided you rely on length < 128. There is only one check if it is greater than MAX_SIFS_FRAME_SIZE.

src/nrf_802154_pib.c
```diff
@@ -63,21 +63,29 @@ typedef struct
...
+    uint8_t                 short_addr[SHORT_ADDRESS_SIZE];         ///< Short Address of this node.
+    uint8_t                 extended_addr[EXTENDED_ADDRESS_SIZE];   ///< Extended Address of this node.
+    nrf_802154_cca_cfg_t    cca;                                    ///< CCA mode and thresholds.
+    bool                    promiscuous : 1;                        ///< Indicating if radio is in promiscuous mode.
+    bool                    auto_ack    : 1;                        ///< Indicating if auto ACK procedure is enabled.
+    bool                    pan_coord   : 1;                        ///< Indicating if radio is configured as the PAN coordinator.
+    uint8_t                 channel     : 5;                        ///< Channel on which the node receives messages.
+    nrf_802154_pib_coex_t   coex;                                   ///< Coex-related fields.
+    nrf_802154_pib_csmaca_t csmaca;                                 ///< CSMA-CA related fields.
+    nrf_802154_pib_ifs_t    ifs;                                    ///< IFS-related fields.
```
ankuns:
```c
#if NRF_802154_IFS_ENABLED
nrf_802154_pib_ifs_t    ifs; 
#endif
```
This would save ram when `!NRF_802154_IFS_ENABLED`

Then in `nrf_802154_pib_ifs_mode_set` allow only value DISABLED, remaining should return false.
Do similar in other ifs related pib functions (return constant when get, disallow write).

src/mac_features/nrf_802154_ifs.c
```diff
@@ -0,0 +1,229 @@
...
+        return true;
+    }
+
+    if (!m_last_frame_length)
+    {
+        // No frame was transmitted before - skip the routine.
+        return true;
+    }
+
+    if (mode == NRF_802154_IFS_MODE_MATCHING_ADDRESSES && !is_ifs_needed_by_address(p_frame))
```
ankuns: Use brackets please `(mode == NRF_802154_IFS_MODE_MATCHING_ADDRESSES)`

src/mac_features/nrf_802154_ifs.c
```diff
@@ -0,0 +1,229 @@
...
+
+    nrf_802154_timer_sched_add(&m_timer, true);
+
+    return false;
+}
+
+void nrf_802154_ifs_transmitted_hook(const uint8_t * p_frame)
+{
+    m_last_frame_timestamp = nrf_802154_timer_sched_time_get();
+
```
ankuns: Here you could assert
```c
assert(p_frame[0] != 0U)
```
You rely on this value, see line 198, and 185.

src/nrf_802154_core_hooks.c
czeslawmakarski: It is placed for a reason (file `nrf_802154_core_hooks.c`, line 59):
```
/* Since some compilers do not allow empty initializers for arrays with unspecified bounds,
 * NULL pointer is appended to below arrays if the compiler used is not GCC. It is intentionally
 * unused, but it prevents the arrays from being empty. GCC manages to optimize empty arrays away,
 * so such a solution is unnecessary. */
```

src/mac_features/nrf_802154_ifs.c
czeslawmakarski: Okay, shall remove.

src/mac_features/nrf_802154_ifs.c
czeslawmakarski: Valid point.

src/mac_features/nrf_802154_ifs.c
czeslawmakarski: Thanks, good point.

src/nrf_802154_core_hooks.c
ankuns: Thanks, I wasn't aware of that. Let's leave it as is.

src/nrf_802154_pib.c
czeslawmakarski: I think it's better to postpone this change (create a Jira task and put it in a backlog), because it is not relevant to IFS alone.

There are lots of conditional functionalities in the driver (i.e. csma-ca, coex etc) which store their configuration in the PIB unconditionally (right now). The point of the task would be to trim down the size and API of PIB depending on compiled-in features. What do you think?

ankuns: Agree, please create an issue for that.

PR #310, opened 2019-11-13
Implementation of Sliding Window CCA feature

czeslawmakarski:
To be switched on/off in the MAC PIB.

### DIFF
--- a/src/nrf_802154_peripherals.h
+++ b/src/nrf_802154_peripherals.h
@@ -299,6 +299,19 @@ extern "C" {
 #define NRF_802154_PPI_RADIO_CRCOK_TO_PPI_GRP_DISABLE NRF_PPI_CHANNEL10
 #endif
 
+/**
+ * @def NRF_802154_PPI_RADIO_CCABUSY_TO_RADIO_CCASTART
+ *
+ * The PPI channel that connects RADIO_CCCABUSY event with the RADIO_CCASTART task.
+ * It is used in case of sliding window option is enabled in the CCA configuration.
+ *
+ * @note This option is used by the core module regardless of the driver configuration.
+ *
+ */
+#ifndef NRF_802154_PPI_RADIO_CCABUSY_TO_RADIO_CCASTART
+#define NRF_802154_PPI_RADIO_CCABUSY_TO_RADIO_CCASTART NRF_PPI_CHANNEL10
+#endif
+
 #if NRF_802154_DISABLE_BCC_MATCHING
 
 /**
--- a/src/nrf_802154_peripherals.h
+++ b/src/nrf_802154_peripherals.h
@@ -505,6 +505,7 @@ extern "C" {
                                            (1 << NRF_802154_PPI_RADIO_CCAIDLE_TO_FEM_GPIOTE) |      \
                                            (1 << NRF_802154_PPI_TIMER_COMPARE_TO_RADIO_TXEN) |      \
                                            (1 << NRF_802154_PPI_RADIO_CRCOK_TO_PPI_GRP_DISABLE) |   \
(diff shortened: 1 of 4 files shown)


src/nrf_802154_peripherals.h
```diff
@@ -299,6 +299,19 @@ extern "C" {
...
+ * @def NRF_802154_PPI_RADIO_CCABUSY_TO_RADIO_CCASTART
+ *
+ * The PPI channel that connects RADIO_CCCABUSY event with the RADIO_CCASTART task.
+ * It is used in case of sliding window option is enabled in the CCA configuration.
+ *
+ * @note This option is used by the core module regardless of the driver configuration.
+ *
+ */
+#ifndef NRF_802154_PPI_RADIO_CCABUSY_TO_RADIO_CCASTART
+#define NRF_802154_PPI_RADIO_CCABUSY_TO_RADIO_CCASTART NRF_PPI_CHANNEL10
```
ankuns: Please update also `NRF_802154_PPI_CHANNELS_USED_MASK`

Is this detected by check in nrf_802154_trx.c line 94 ?

src/nrf_802154_trx.c
```diff
@@ -1504,15 +1549,28 @@ void nrf_802154_trx_standalone_cca(void)
 {
     assert((m_trx_state == TRX_STATE_IDLE) || (m_trx_state == TRX_STATE_FINISHED));
 
+    bool sliding_window = is_sliding_window_enabled();
+
     m_trx_state = TRX_STATE_STANDALONE_CCA;
 
     // Set shorts
-    nrf_radio_shorts_set(SHORTS_CCA);
+    nrf_radio_shorts_set(sliding_window ? SHORTS_CCA_SLIDING : SHORTS_CCA);
```
ankuns: Sliding window in standalone cca is not defined in requirements of the feature. In my opinion standalone cca should not support sliding window feature. Please consult @hubertmis . What would be the reason of such function. Standalone cca is only to detect if channel is busy or not. With sliding window enabled, behavior changes to "wait until channel is idle", and the result is meaningless.

src/nrf_802154_trx.c
```diff
@@ -1530,6 +1588,7 @@ static void standalone_cca_finish(void)
 {
     nrf_ppi_channel_disable(PPI_DISABLED_EGU);
     nrf_ppi_channel_disable(PPI_EGU_RAMP_UP);
+    nrf_ppi_channel_disable(PPI_CCABUSY_CCASTART);
```
ankuns: If we reach consensus about not using sliding window feature in nrf_802154_trx_standalone_cca this will be removed.

src/nrf_802154_trx.c
hubertmis: Some implementations use standalone CCA to implement CSMA-CA in the layer above the driver. I think if such implementation selected the sliding window CCA, it would expect it both in CCA before transmission and in standalone CCA. That's why I recommended to implement it in both.
If someone would like to use normal CCA during standalone procedure, it should be ok to configure it run-time when the transceiver is disabled.

ankuns: Let it be so.

src/nrf_802154_trx.c
ankuns: Not the case anymore as we have consensus.

src/nrf_802154_peripherals.h
czeslawmakarski: Sure, I'll update. It was not detected by the check because we reuse the same PPI channels, so in the logical OR there are 2 same PPIs, which is a natural thing, since we would like to be conservative on the number of used PPIs.

PR #311, opened 2019-11-14
Added 802.15.4 AES-CCM* encryption option

mikaczma:
Added encryption module for 802.15.4 driver.
It is designed according to IEEE 802.15.4-2016 standard.


hubertmis:
Thanks for this contribution!
Please, rebase your branch to the current upstream master and use `./scripts/pretty.sh` to make sure the code is compliant with the coding standard.
We'll take care of the continuous integration servers to make sure bamboo lets us merge.

src/nrf_802154_encrypt.h
```diff
@@ -0,0 +1,107 @@
...
+ */
+void nrf_802154_encrypt_schedule_aes_ccm_auth_transform(const nrf_802154_aes_ccm_frame_t * p_frame);
+
+/**
+ * @brief Callback to tx_started notification
+ *        During this callback transformation is started
+ * 
+ * @param[in] pointer to raw transmitted frame
+ */
+void nrf_802154_tx_started(const uint8_t * p_frame);
```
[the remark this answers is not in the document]
mikaczma: this name is inherited from extern weak callback from nrf_802154.c file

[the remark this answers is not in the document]
ankuns: Please consult @hubertmis , I think this callout is provided to the application. We should find another way of triggering encryption on transmit start, for example by some transmit hooks.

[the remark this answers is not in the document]
hubertmis: For now I suggested to do that as it is implemented. I think we'll have to create a "core hook" to handle it properly. In current intermediate solution, the user of the driver can use this encryption module what disables possibility to get `nrf_802154_tx_started` notification in the application, or disable this module and use the notification.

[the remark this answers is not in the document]
ankuns: This will be technical debt. Let it be.

src/nrf_802154_encrypt.c
```diff
@@ -0,0 +1,481 @@
+/* Copyright (c) 2015 Nordic Semiconductor ASA and Luxoft Global Operations Gmbh.
```
jciupis: Is the license up to date?

src/nrf_802154_encrypt.c
```diff
@@ -0,0 +1,481 @@
...
+#define CYPHERTEXT (nrf_ecb_data_pointer_get(NRF_ECB) + 32)                            ///< Macro AES CipherText Address
+
+static nrf_802154_encrypt_aes_ccm_frame_t m_aes_ccm_frame;                             ///< AES CCM Frame
+static uint8_t                            m_auth_key[NRF_802154_ENCRYPT_BLOCK_SIZE];   ///< Stored AES Key
+static uint8_t                            m_auth_nonce[NRF_802154_ENCRYPT_NONCE_SIZE]; ///< Stored AES-CCM* Nonce
+static uint8_t                            m_x[NRF_802154_ENCRYPT_BLOCK_SIZE];          ///< CBC-MAC value - Annex B4.1.2 d)
+static uint8_t                            m_b[NRF_802154_ENCRYPT_BLOCK_SIZE];          ///< B[i] octet for Authorization Transformatino - Annex B4.1.2 b)
+static uint8_t                            m_m[NRF_802154_ENCRYPT_BLOCK_SIZE];          ///< M[i] octet as parsed plaintext blocks - Annex B4.1.3 c)
+static uint8_t                            m_a[NRF_802154_ENCRYPT_BLOCK_SIZE];          ///< A[i] octet for Encryption Transformation - Annex B4.1.3 b)
+static ccm_state_t                        m_state;;                                    ///< State of AES-CCM* transformation
```
jciupis: typo: double semicolon

src/nrf_802154_encrypt.c
```diff
@@ -0,0 +1,481 @@
...
+    uint8_t l;
+
+    auth_flags |= (p_frame->auth_data_len == 0) ? 0 : NRF_802154_ENCRYPT_ADATA_AUTH_FLAG;
+
+    m           = m_mic_size[p_frame->mic_level];
+    m           = (m > 0) ? (m - 2) >> 1 : 0;
+    auth_flags |= (m << NRF_802154_ENCRYPT_M_BITS_AUTH_FLAG);
+
+    l           = NRF_802154_ENCRYPT_L_VALUE - 1;
+    auth_flags |= l;
```
jciupis: Variable named `l` is so easy to confuse with a `1` that it should either be renamed or removed completely, which I believe could be the case here:
`auth_flags |= (NRF_802154_ENCRYPT_L_VALUE - 1);`

src/nrf_802154_encrypt.c
```diff
@@ -0,0 +1,481 @@
...
+
+    return auth_flags;
+}
+
+/**
+ * @brief   Forms additional authentic data from octet string a by 16-octet chunks
```
jciupis: `authenticated`? `authentication`? I assume `authentic` is not the word you were looking for here ;)

src/nrf_802154_encrypt.c
```diff
@@ -0,0 +1,481 @@
...
+ * IEEE std 802.15.4-2015, B.4.1.1 Input transformation
+ *
+ * @param[in]  p_frame pointer to AES CCM frame structure
+ * @param[in]  iter    number of chunk
+ * @param[out] p_b     pointer to memory for Bi
+ *
+ * @return true - chunk was formated, false - otherwise
```
jciupis: Please use the `@retval` syntax:
```
@retval true   Condition satisfied.
@retval false  Otherwise.
```

src/nrf_802154_encrypt.c
```diff
@@ -0,0 +1,481 @@
...
+    {
+        return false;
+    }
+
+    memset(p_b, 0, NRF_802154_ENCRYPT_BLOCK_SIZE);
+
+    if (iter == 0)
+    {
+        len    = MIN(p_frame->auth_data_len, NRF_802154_ENCRYPT_BLOCK_SIZE - sizeof(uint16_t));
+        p_b[0] = (p_frame->auth_data_len & 0xFF00) >> 8;
```
jciupis: It'd be cool to get rid of as many magic numbers as possible. While the bit shifts are straightforward enough, I'd suggest replacing the array indices with self-explaining macros and use it in other functions (e.g. `b0_format`) as well.

src/nrf_802154_encrypt.c
```diff
@@ -0,0 +1,481 @@
...
+
+/**
+ * @brief   Forms plain/cipher text data from octet string m/c by 16-octet chunks
+ * IEEE std 802.15.4-2015, B.4.1.1 Input transformation
+ *
+ * @param[in]  p_frame pointer to AES CCM frame structure
+ * @param[in]  iter    number of chunk
+ * @param[out] p_b     pointer to memory for Bi
+ *
+ * @return true - chunk was formated, false - otherwise
```
jciupis: Same as 179

src/nrf_802154_encrypt.c
```diff
@@ -0,0 +1,481 @@
...
+    memcpy(CLEARTEXT, m_a, NRF_802154_ENCRYPT_BLOCK_SIZE);
+    NRF_ECB->TASKS_STARTECB = 1;
+}
+
+/**
+ * @brief helper function for plain text encryption in ECB IRQ
+ */
+static void perform_plain_text_encryption(void)
+{
+    memcpy(m_auth_tag, CYPHERTEXT, m_mic_size[m_aes_ccm_frame.mic_level]);
```
jciupis: Please add newline here and after the `m_state` assignments for readability

src/nrf_802154_encrypt.c
```diff
@@ -0,0 +1,481 @@
...
+ * @param[in]  flags   flags for injection to B0 field
+ * @param[out] p_b     pointer to memory for B0
+ */
+static void b0_format(const nrf_802154_encrypt_aes_ccm_frame_t * p_frame,
+                      const uint8_t                              flags,
+                      uint8_t                                  * p_b)
+{
+    p_b[0] = flags;
+    memcpy(&p_b[1], p_frame->nonce, NRF_802154_ENCRYPT_NONCE_SIZE);
+    p_b[NRF_802154_ENCRYPT_BLOCK_SIZE - 2] = (p_frame->plain_text_data_len & 0xFF00) >> 8;
```
czeslawmakarski: You can make these two lines coherent with the lines 123-124.

src/nrf_802154_encrypt.c
```diff
@@ -0,0 +1,481 @@
...
+    uint8_t m;
+    uint8_t l;
+
+    auth_flags |= (p_frame->auth_data_len == 0) ? 0 : NRF_802154_ENCRYPT_ADATA_AUTH_FLAG;
+
+    m           = m_mic_size[p_frame->mic_level];
+    m           = (m > 0) ? (m - 2) >> 1 : 0;
+    auth_flags |= (m << NRF_802154_ENCRYPT_M_BITS_AUTH_FLAG);
+
+    l           = NRF_802154_ENCRYPT_L_VALUE - 1;
```
czeslawmakarski: `l` seems to be used only here. Why not just `auth_flags |= (NRF_802154_ENCRYPT_L_VALUE - 1)

src/nrf_802154_encrypt.c
```diff
@@ -0,0 +1,481 @@
...
+    {
+        memcpy(&m_auth_nonce, p_nonce, NRF_802154_ENCRYPT_NONCE_SIZE);
+        m_aes_ccm_frame.nonce = m_auth_nonce;
+    }
+}
+
+void nrf_802154_encrypt_schedule_aes_ccm_auth_transform(
+    const nrf_802154_encrypt_aes_ccm_frame_t * p_frame)
+{
+    if ((p_frame->auth_data != NULL) && (p_frame->key != NULL)
```
czeslawmakarski: Why don't you turn this one huge `if()` clause into series of `if(!)` clauses?
i.e.
`if (p_frame->auth_data == NULL)
{
    return;
}

if (p_frame_key == NULL)
{
    return;
}
```
and so on?

hubertmis:
Looks good :+1: 
Have you signed CLA? It looks that CLAassistant cannot verify it.

PR #312, opened 2019-11-14
Refactor timeslot dropping procedure

jciupis:
This PR introduces initial changes needed to bring the multi-protocol solution back from the dead.

This is still in experimental stage and more fixes might (and most probably will) follow.

### DIFF
--- a/src/nrf_802154_core.c
+++ b/src/nrf_802154_core.c
@@ -1119,7 +1119,10 @@ void nrf_802154_rsch_crit_sect_prio_changed(rsch_prio_t prio)
         on_timeslot_ended();
         return;
     }
-    // TODO: how to motivate adding this else-if?
+    // It might happen that even though IDLE has already been notified, this function is called
+    // again as a result of preemptions caused by unexpected timeslot change (e.g. the application
+    // requested transition to sleep while out of timeslot and RAAL notified timeslot start
+    // in the middle of that sleep request). The following block makes RAAL finish its processing.
     else if (prio == RSCH_PRIO_IDLE)
     {
         nrf_802154_rsch_continuous_ended();
--- a/src/nrf_802154_core.c
+++ b/src/nrf_802154_core.c
@@ -712,20 +712,12 @@ static bool current_operation_terminate(nrf_802154_term_t term_lvl,
 /** Enter Sleep state. */
 static void sleep_init(void)
 {
-    nrf_802154_trx_disable();
     nrf_802154_timer_coord_stop();
 }
 
 /** Initialize Falling Asleep operation. */
 static void falling_asleep_init(void)
 {
-    if (!timeslot_is_granted())
-    {
-        sleep_init();
-        state_set(RADIO_STATE_SLEEP);
-        return;
-    }
-
     if (nrf_802154_trx_go_idle())
     {
         // There will be nrf_802154_trx_in_idle call, where we will continue processing
@@ -951,6 +943,8 @@ static void on_timeslot_ended(void)
 
         nrf_802154_timer_coord_stop();
 
+        nrf_802154_rsch_continuous_ended();
+
         result = nrf_802154_core_hooks_terminate(NRF_802154_TERM_802154, REQ_ORIG_RSCH);
         assert(result);
         (void)result;
@@ -1106,9 +1100,9 @@ void nrf_802154_rsch_crit_sect_prio_changed(rsch_prio_t prio)
 
     m_rsch_priority = prio;
 
+    // We have just got a timeslot.
     if ((old_prio == RSCH_PRIO_IDLE) && (prio != RSCH_PRIO_IDLE))
     {
-        // We have just got a timeslot.
         nrf_802154_log(EVENT_TRACE_ENTER, FUNCTION_TIMESLOT_STARTED);
 
         nrf_802154_trx_enable();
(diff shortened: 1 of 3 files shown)


src/nrf_802154_core.c
```diff
@@ -1106,9 +1100,9 @@ void nrf_802154_rsch_crit_sect_prio_changed(rsch_prio_t prio)
 
     m_rsch_priority = prio;
 
+    // We have just got a timeslot.
```
czeslawmakarski: I know it's a matter of choice, but I always placed such commits as a first line inside the scope, i.e.:
```
if ((old_prio == RSCH_PRIO_IDLE) && (prio != RSCH_PRIO_IDLE))
{
    // We have just got a timeslot.
```

src/rsch/raal/softdevice/nrf_raal_softdevice.c
```diff
@@ -163,34 +172,26 @@ static timer_action_t m_timer_action;
 /**@brief Current timeslot length. */
 static uint16_t m_timeslot_length;
 
-/**@brief Previously granted timeslot length. */
+/**@brief Previously granted timeslot length.
+ *        The RAAL timer uses this value to fire by the end of the previously granted timeslot
```
ankuns: Whitespace

src/nrf_802154_core.c
```diff
@@ -1748,8 +1754,16 @@ bool nrf_802154_core_sleep(nrf_802154_term_t term_lvl)
 
             if (result)
             {
-                state_set(RADIO_STATE_FALLING_ASLEEP);
-                falling_asleep_init();
+                if (!timeslot_is_granted())
```
czeslawmakarski: Why don't you switch the branches and omit the `!`?

src/nrf_802154_core.c
ankuns: There are also other places written in this style.

jciupis: I have no strong opinion on where such comments should be placed. I'll align all the changes with your suggestion

src/rsch/raal/softdevice/nrf_raal_softdevice.c
```diff
@@ -485,135 +521,231 @@ static void timer_irq_handle(void)
...
-            nrf_802154_log(EVENT_TRACE_ENTER, FUNCTION_RAAL_SIG_EVENT_EXTEND_FAIL);
+    timeslot_state_set(TIMESLOT_STATE_IDLE);
 
-            if (!timer_is_set_to_margin())
-            {
-                timer_to_margin_set();
-            }
+    if (m_continuous)
+    {
+        if (m_timeslot_extend_tries < m_config.sd_cfg.timeslot_alloc_iters)
```
hubertmis: If I remember correctly, we should remove this whole `if` block to not decrease length of requested timeslot on `BLOCKED` event. We should request always 6400us on `BLOCKED`.
But it's out of scope of this task. We can create a Jira to do that

src/rsch/raal/softdevice/nrf_raal_softdevice.c
```diff
@@ -804,14 +905,27 @@ void nrf_raal_continuous_mode_exit(void)
 
 void nrf_raal_continuous_ended(void)
 {
-    // Intentionally empty.
+    // Timeslot was dropped intentionally
```
hubertmis: Is it possible to enter this function if timeslot is not dropped intentionally? If safety margin is the case, I think we should add it to this comment.

src/rsch/raal/softdevice/nrf_raal_softdevice.c
jciupis: If we're certain that it's the desirable behaviour, then I can change it straight away, otherwise there'll be more time spent on creating tickets than removing the block ;) @ankuns @czeslawmakarski, do you agree on requesting the full timeslot when blocked?

src/rsch/raal/softdevice/nrf_raal_softdevice.c
jciupis: I really need to put comments inside the blocks as @czeslawmakarski suggested! ;) This comment is related to the following case, not to the function in general. The design assumes that this function is entered **every time** the timeslot stops being controlled by 802.15.4, whatever the reason.

hubertmis: I understood it's for the `if` block. Just wasn't sure if can be replaced with `assert` - I think it should be explained here when do we expect that this function is called and state is other than `DROPPED`

jciupis: In the current implementation, we only expect this function to be called in two cases:
* the timeslot was revoked by the radio arbiter and at the moment of this function's execution it is already `IDLE`;
* the timeslot was dropped intentionally by the higher layer and it is `DROPPED`.

Asserting might seem a reasonable idea then, but I'm against asserting on the timeslot state here. Although it is not currently possible, the design allows for a precondition other than RAAL to revoke access to the RADIO peripheral, that would result in core being notified about approved preconditions priority equal `IDLE`. In that case, this function would be called with RAAL being e.g. `GRANTED` and I think we should drop the notification then, just like the introduced patch does right now.

Definitely worth a comment with explanation though!

hubertmis: Thanks for the detailed explanation. I understand now and agree `assert` is not the option here, at least in the current design.
I would suggest to add more details to the comment.

PR #314, opened 2019-11-18
Pr/clean up logging

ankuns:
Changed logging method

To reviewers:
Please don't review decoder.html yet, this is still under development (I'll push and re-request review).


src/mac_features/nrf_802154_csma_ca.c
```diff
@@ -153,7 +167,7 @@ static void frame_transmit(rsch_dly_ts_id_t dly_ts_id)
         }
     }
 
-    nrf_802154_log(EVENT_TRACE_EXIT, FUNCTION_CSMA_FRAME_TRANSMIT);
+    nrf_802154_debug_log_function_exit(2);
```
jciupis: The verbosity level of this function's `exit` is higher than that of `entry`. Is it intentional?

src/nrf_802154_trx.c
```diff
@@ -1877,27 +2063,38 @@ static void irq_handler_crcok(void)
...
 static void txframe_finish_disable_ppis(void)
 {
+    nrf_802154_debug_log_function_entry(2);
+
     nrf_ppi_channel_disable(PPI_DISABLED_EGU);
     nrf_ppi_channel_disable(PPI_EGU_RAMP_UP);
     nrf_ppi_fork_endpoint_setup(PPI_EGU_RAMP_UP, 0);
     nrf_ppi_channel_remove_from_group(PPI_EGU_RAMP_UP, PPI_CHGRP0);
     nrf_ppi_channel_disable(PPI_CCABUSY_CCASTART);
+    nrf_802154_debug_log_function_exit(2);
```
czeslawmakarski: Newline missing.

src/rsch/nrf_802154_rsch.c
```diff
@@ -429,11 +436,14 @@ static bool precise_delayed_timeslot_request(dly_ts_t                  * p_dly_t
...
         nrf_802154_timer_sched_add(&p_dly_ts->timer, true);
 
+        nrf_802154_debug_log_function_exit(1);
+
         return true;
     }
     else
     {
         // The requested time is in the past.
+        nrf_802154_debug_log_function_exit(1);
```
czeslawmakarski: Newline.

src/rsch/nrf_802154_rsch.c
```diff
@@ -411,6 +417,7 @@ static bool precise_delayed_timeslot_request(dly_ts_t                  * p_dly_t
 
         nrf_802154_timer_sched_add(&p_dly_ts->timer, false);
 
+        nrf_802154_debug_log_function_exit(1);
```
czeslawmakarski: Newline.

src/mac_features/nrf_802154_csma_ca.c
```diff
@@ -250,6 +266,8 @@ void nrf_802154_csma_ca_start(const uint8_t * p_data)
     m_is_running = true;
 
     random_backoff_start();
```
jciupis: Shouldn't `random_backoff_start` function be logged with the higher verbosity value (`2`)?

tools/event_decoder/make_addr2fname_map.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# Tool for creating maping:
+# - from address of string created by using __func__ special identifier of C language
+# - to content of the string
+
+if [ "$1" == "" ]; then
+	echo "Usage:"
+	echo "make_addr2fname_map.sh <input_elf_file>"
+	echo ""
+	echo "input_elf_file    The name (with path if necessary) to elf file to parse"
```
czeslawmakarski: I would omit the `(with path if necessary)`. It is natural that the file name may contain an absolute or relative path.

tools/event_decoder/make_addr2fname_map.sh
```diff
@@ -0,0 +1,48 @@
...
+
+INPUT_ELF_FILE=$1
+
+if [ ! -f "$INPUT_ELF_FILE" ]; then
+    echo "'$INPUT_ELF_FILE' does not exist"
+	exit 1
+fi
+
+# Prepare command script for gdb (initial part)
+echo > make_addr2fname_map_script.gdb
```
czeslawmakarski: I would substitute `make_addr2fname_map_script.gdb` with `/tmp/make_addr2fname_map_script.gdb`, so the script does not leave anything in the CWD.

src/mac_features/nrf_802154_delayed_trx.c
```diff
@@ -391,6 +399,8 @@ static void rx_timeslot_started_callback(rsch_dly_ts_id_t dly_ts_id)
             assert(false);
             break;
     }
+
+    nrf_802154_debug_log_function_exit(2);
 }
 
 bool nrf_802154_delayed_trx_transmit(const uint8_t * p_data,
```
jciupis: I think more functions in this module should be logged. They can all follow a high verbosity setting so that it needs to be set explicitly to overwrite the default setting. Is there any particular reason for not logging them (e.g. redundancy)?

src/nrf_802154_debug_log.h
```diff
@@ -0,0 +1,212 @@
...
+ *                          of the parameter is defined by the module in which
+ *                          the log is recorded and event_id.
+ */
+#define nrf_802154_debug_log_local_event(verbosity, local_event_id, param_u16)  \
+do                                                                              \
+{                                                                               \
+    if (nrf_802154_debug_log_verbosity_allows(verbosity))                       \
+    {                                                                           \
+        nrf_802154_debug_log_write_raw(                                         \
+                ((NRF_802154_DEBUG_LOG_TYPE_LOCAL_EVENT) << 28) |               \
```
jciupis: The magic numbers `28` and `22` should be replaced with macros and documented in their definitions. No need to list possible values in the function documentation then

src/nrf_802154_debug_log.h
```diff
@@ -0,0 +1,212 @@
...
+ */
+#ifndef NRF_802154_DEBUG_LOG_BUFFER_LEN
+#define NRF_802154_DEBUG_LOG_BUFFER_LEN 1024U
+#endif
+
+/**@def NRF_802154_DEBUG_LOG_BLOCKS_IRQS
+ * @brief Configures if writing to log buffer is performed with interrupts disabled.
+ * Setting this macro to 1 has following consequences:
+ * - Interrupts are automatically disabled during write to log buffer. This ensures
+ *   thread-safety and always coherent log)
```
jciupis: typo: unnecessary `)`

src/nrf_802154_debug_log.c
```diff
@@ -34,18 +34,20 @@
...
+
 /// Index of the log buffer pointing to the element that should be filled with next log message.
 volatile uint32_t nrf_802154_debug_log_ptr = 0;
 
 void nrf_802154_debug_log_init(void)
 {
-    // Intentionally empty
+    // Let's check if address of generated __func__ will fit into log generated
+    // by @ref nrf_802154_debug_log_function_entry
+    assert((uint32_t)((uintptr_t)(__func__)) < (1U << 22)) ;
```
jciupis: Please replace the magic number with a macro

tools/event_decoder/make_addr2fname_map.sh
```diff
@@ -0,0 +1,48 @@
...
+echo set pagination off >> make_addr2fname_map_script.gdb
+echo set height unlimited >> make_addr2fname_map_script.gdb
+
+# Prepare command script for gdb (commands reading content of elf file at addresses we are interested in)
+# Steps (according to pipe)
+# 1. Using readelf read sybmol table of input elf file
```
jciupis: typo: `symbol`

src/rsch/nrf_802154_rsch.c
```diff
@@ -202,6 +205,7 @@ static inline void all_prec_update(void)
     {
         if (!mutex_trylock(&m_req_mutex, &m_req_mutex_monitor))
         {
+            nrf_802154_debug_log_function_exit(2);
```
jciupis: Why not break out of the loop and use the existing log exit instead?

src/rsch/nrf_802154_rsch.c
```diff
@@ -429,11 +436,14 @@ static bool precise_delayed_timeslot_request(dly_ts_t                  * p_dly_t
         all_prec_update();
         nrf_802154_timer_sched_add(&p_dly_ts->timer, true);
 
+        nrf_802154_debug_log_function_exit(1);
```
jciupis: Please replace `return` with assignment to a bool variable and provide a single point of exit with a single log.

src/mac_features/nrf_802154_delayed_trx.c
ankuns: The reason here was to log where it logged before, I'll add more logs

tools/event_decoder/debugLogDecoder.js
```diff
@@ -0,0 +1,174 @@
...
+        switch (event.paramType) 
+        {
+            case "uint":
+                result += eventParam.toString();
+                break;
+            case "enum":
+                result += this.enumValueToText(event.enumValues, eventParam);
+                break;
+            default:
+				// Possibly you have typo or introduced new paramType value not handled here
```
jciupis: Some crazy whitespace here

tools/event_decoder/debugLogDecoder.js
```diff
@@ -0,0 +1,174 @@
...
+        result = this.createUnknownFunctionObj(functionId);
+    }
+    return result;
+}
+
+DebugLogDecoder.prototype.parseEventCode = function(eventCode)
+{
+    var result = {};
+    
+    result.eventType = (eventCode >> 28) & 0xF;
```
jciupis: Is it possible to use the same or similar names for the magic numbers here as in the C code? Would make it easier not to forget to change these values when modifying them in the C code

tools/event_decoder/debugLogDecoder.js
```diff
@@ -0,0 +1,174 @@
...
+            break;
+            
+        default:
+            result.module = this.createUnknownModule(0);
+            result.text = "Unknown event type (" + eventCode + ")";
+            break;
+    }
+    
+    return result;
+}
```
jciupis: I'd add a newline at the end of file, just in case

tools/event_decoder/debugLogDecoderData.js
```diff
@@ -0,0 +1,113 @@
...
+// please update this file.
+
+
+// Description of modules that may be present in logs
+// Each member of the array is an object with following fields:
+// id (int) - Module id, set up in C source code by NRF_802154_DEBUG_LOG_MODULE_ID
+// name (string) - Name of the module to be displayed on sequence diagram
+// localEvents (array of objects or undefined) - local events that may be logged by the module. This is array of objects with following fields:
+//    id (int) - identifier of an event (See C defines NRF_802154_DEBUG_LOG_LOCAL_EVENT_ID_<MODULE_NAME>_xxxx
+//    text (string) - humen readable text describing an event (will be visible in sequence diagram)
```
jciupis: typo: `human`

tools/event_decoder/decoder.html
```diff
@@ -60,341 +60,119 @@ <h3 class="text-center">Sequence diagram generator for <span class="label label-
...
-            this.input = "";
-
-            // Transform GDB input to JSON compatible object.
-            debug_json = debug_json.substring(debug_json.indexOf("{"));
-            debug_json = debug_json.replace(/0x[0-9a-fA-F]{8}/g, function myFunction(x){return '"' + x + '"'});
-            debug_json = debug_json.replace('{', '[');
-            debug_json = debug_json.replace('}', ']');
+            str = str.replace('{', '[');
+            str = str.replace('}', ']');
+            str = str.replace(/0x[0-9a-fA-F]{8}/g, function myFunction(x){return parseInt(x)});
```
jciupis: I know this code will not be read frequently, but `myFunction` could be replaced with something more meaningful :)

src/nrf_802154_debug_log.c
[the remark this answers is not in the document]
ankuns: ad. 1. This assert may pass while other `__func__` addresses could fail to fit into 22-bit space. I have no good solution. Checking at `nrf_802154_debug_log_function_entry` would introduce runtime overhead. Asserting here just makes things a bit simpler if someone tweaks liner script in such a way that whole .rodata gets into higher addresses.

What would you advice?

ad. 2. I agree, I'll introduce `#define`

tools/event_decoder/make_addr2fname_map.sh
ankuns: I agree, but on windows gdb is unable to find `/tmp/make_addr2fname_map_script.gdb`, even from git-bash, I'll let it as is.

tools/event_decoder/debugLogDecoder.js
ankuns: done

PR #317, opened 2019-11-21
Int/wifi

mikaczma:
Added CSL injection feature


src/mac_features/nrf_802154_frame_parser.c
```diff
@@ -680,3 +690,83 @@ const uint8_t * nrf_802154_frame_parser_ie_header_get(const uint8_t * p_frame)
...
+    if (p_ie_header == NULL)
+    {
+        return false;
+    }
+
+    while ((nrf_802154_frame_parser_ie_header_type_get(p_ie_header) !=
+            IE_HEADER_TERMINATION_1_ID) &&
+           (nrf_802154_frame_parser_ie_header_type_get(p_ie_header) != IE_HEADER_TERMINATION_2_ID))
+    {
+        switch (nrf_802154_frame_parser_ie_header_type_get(p_ie_header))
```
ankuns: If there are no plans for extending this switch I would recommend using just `if` instead of `switch`

src/nrf_802154_csl.c
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * @brief   Update CSL Phase field
+ * IEEE std 802.15.4-2015, B.4.1.2 Authentication transformation
+ *
+ * @param[in] p_frame Pointer to IE CSL header
+ * @param[in] phase   Value of actual CSL phase given in symbols
+ */
+static void nrf_802154_csl_phase_update(const uint8_t * p_ie_csl_header, uint16_t phase)
+{
+    *(uint16_t *)(p_ie_csl_header + IE_CSL_HEADER_PHASE_BYTE) = phase / 10;
```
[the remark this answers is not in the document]
ankuns: If we need to port to CPU that doesn't support unaligned access (e.g. Cortex-M0) there we will have mysterious fails. Please don't cast from `uint8_t *` to pointer pointing to wider types unless you are absolutely sure there will be no alignment issues.

src/mac_features/nrf_802154_frame_parser.c
```diff
@@ -680,3 +690,83 @@ const uint8_t * nrf_802154_frame_parser_ie_header_get(const uint8_t * p_frame)
...
 }
+
+const uint8_t nrf_802154_frame_parser_ie_header_length_get(const uint8_t * p_ie_header)
+{
+    if (p_ie_header == NULL)
+    {
+        return 0;
+    }
+
+    return (uint8_t)(*((const uint16_t *)p_ie_header) & 0x7F);
```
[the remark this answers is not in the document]
ankuns: Potential alignment problems, casting pointer `uint8_t*` to pointer to wider type.

src/mac_features/nrf_802154_frame_parser.c
```diff
@@ -680,3 +690,83 @@ const uint8_t * nrf_802154_frame_parser_ie_header_get(const uint8_t * p_frame)
...
+}
+
+const uint8_t nrf_802154_frame_parser_ie_header_type_get(const uint8_t * p_ie_header)
+{
+    if (p_ie_header == NULL)
+    {
+        return 0;
+    }
+
+    return (uint8_t)(*((const uint16_t *)p_ie_header) >> 7);
```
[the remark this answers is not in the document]
ankuns: Potential alignment problems, casting pointer `uint8_t*` to pointer to wider type.

src/rsch/nrf_802154_rsch.c
```diff
@@ -570,3 +572,13 @@ void nrf_802154_wifi_coex_denied(void)
...
+
+uint32_t nrf_802154_rsch_get_next_scheduled_receive_time(void)
+{
+    return m_dly_ts[RSCH_DLY_RX].t0 + m_dly_ts[RSCH_DLY_RX].dt;
+}
+
+uint32_t nrf_802154_rsch_get_next_scheduled_receive_symbols(void)
+{
+    return nrf_802154_rsch_get_next_scheduled_receive_time() * 1000000 / SYMBOLS_PER_SECOND;
+}
```
ankuns: Missing newline

src/nrf_802154_csl.c
```diff
@@ -0,0 +1,75 @@
...
+}
+
+void nrf_802154_tx_started(const uint8_t * p_frame)
```
ankuns: Function with this name is implemented in `nrf_802154_encrypt.c`.

src/nrf_802154_encrypt.c
```diff
@@ -0,0 +1,526 @@
...
+}
+
+#ifdef ENCRYPT_TX_STARTED
+
+void nrf_802154_tx_started(const uint8_t * p_frame)
+{
+    nrf_802154_encrypt_aes_ccm_auth_transform_trigger(p_frame);
+}
+
+#endif // ENCRYPT_TX_STARTED
```
ankuns: missing newline

src/nrf_802154_csl.c
```diff
@@ -0,0 +1,75 @@
...
+#define IE_CSL_HEADER_PHASE_BYTE 2 ///< IE CSL Header Phase octet position
+
+/**
+ * @brief   Update CSL Phase field
+ * IEEE std 802.15.4-2015, B.4.1.2 Authentication transformation
+ *
+ * @param[in] p_frame Pointer to IE CSL header
+ * @param[in] phase   Value of actual CSL phase given in symbols
+ */
+static void nrf_802154_csl_phase_update(const uint8_t * p_ie_csl_header, uint16_t phase)
```
ankuns: I think prototype should be:
`static void nrf_802154_csl_phase_update(uint8_t * p_ie_csl_header, uint16_t phase)`
We should move cast removing `const` to more general place like, `nrf_802154_tx_started` and **comment** there that by design we update what's behind `const uint8_t * p_frame` pointer in-place. We will need further rework to make it const-clean.

src/mac_features/nrf_802154_frame_parser.h
```diff
@@ -314,4 +325,48 @@ const uint8_t * nrf_802154_frame_parser_ie_header_get(const uint8_t * p_frame);
...
+/**
+ * @brief Gets length of IE header content
+ *
+ * @param[in] p_ie_header  Pointer to IE header
+ *
+ * @returns Length of IE header content in bytes
+ * @returns Zero if IE header pointer is NULL
+ *
+ */
+const uint8_t nrf_802154_frame_parser_ie_header_length_get(const uint8_t * p_ie_header);
```
jciupis: I think the `const` specifier for the return value can be removed.

src/mac_features/nrf_802154_frame_parser.h
```diff
@@ -314,4 +325,48 @@ const uint8_t * nrf_802154_frame_parser_ie_header_get(const uint8_t * p_frame);
...
+/**
+ * @brief Gets type ID of IE header
+ *
+ * @param[in] p_ie_header  Pointer to IE header
+ *
+ * @returns Type ID of IE header content
+ * @returns Zero if IE header pointer is NULL
+ *
+ */
+const uint8_t nrf_802154_frame_parser_ie_header_type_get(const uint8_t * p_ie_header);
```
jciupis: I think the `const` specifier for the return value can be removed.

src/mac_features/nrf_802154_frame_parser.c
```diff
@@ -680,3 +690,83 @@ const uint8_t * nrf_802154_frame_parser_ie_header_get(const uint8_t * p_frame)
...
+    if (p_ie_header == NULL)
+    {
+        return NULL;
+    }
+
+    while ((nrf_802154_frame_parser_ie_header_type_get(p_ie_header) !=
+            IE_HEADER_TERMINATION_1_ID) &&
+           (nrf_802154_frame_parser_ie_header_type_get(p_ie_header) != IE_HEADER_TERMINATION_2_ID))
+    {
+        switch (nrf_802154_frame_parser_ie_header_type_get(p_ie_header))
```
jciupis: What's the benefit of a `switch` over a simple `if` here?

src/rsch/nrf_802154_rsch.c
```diff
@@ -570,3 +572,13 @@ void nrf_802154_wifi_coex_denied(void)
     prec_approved_prio_set(RSCH_PREC_COEX, RSCH_PRIO_RX);
     notify_core();
 }
+
+uint32_t nrf_802154_rsch_get_next_scheduled_receive_time(void)
+{
+    return m_dly_ts[RSCH_DLY_RX].t0 + m_dly_ts[RSCH_DLY_RX].dt;
```
jciupis: Note that neither this function's implementation nor its callers check if the provided time represents an actually scheduled, pending delayed RX window. Could you have a look into the 802.15.4-2015 spec to determine what should be done with the CSL phase when this time is in the past or happens to be invalid for any other reason (e.g. the delayed timeslot was cancelled)? Currently it silently passes the possibly invalid data to the frame, the spec might have different requirements.

src/mac_features/nrf_802154_frame_parser.c
```diff
@@ -691,24 +686,28 @@ const uint8_t * nrf_802154_frame_parser_ie_header_get(const uint8_t * p_frame)
...
-const uint8_t nrf_802154_frame_parser_ie_header_type_get(const uint8_t * p_ie_header)
+uint8_t nrf_802154_frame_parser_ie_header_type_get(const uint8_t * p_ie_header)
 {
     if (p_ie_header == NULL)
     {
         return 0;
     }
 
-    return (uint8_t)(*((const uint16_t *)p_ie_header) >> 7);
+    uint8_t ie_header_type = (p_ie_header[0] & 0x80) >> 7;
```
jciupis: Please replace as many magic numbers with macros as possible. Examples: `p_ie_header[PHR_OFFSET]` instead of `p_ie_header[0]`, `MAX_PACKET_SIZE` instead of `0x7F`. Macros useful for this purpose can be found in `nrf_802154_const.h`. Please add more if need be.

src/nrf_802154_const.h
```diff
@@ -82,6 +82,11 @@
 #define IE_PRESENT_OFFSET            2                                            ///< Byte containing the IE Present bit.
 #define IE_PRESENT_BIT               0x02                                         ///< Bits containing the IE Present field.
 
+#define IE_HEADER_BYTES              2                                            ///< Length of IE header without content
```
jciupis: `IE_HEADER_LENGTH` name would fit better for `Length of IE header` ;)

src/mac_features/ack_generator/nrf_802154_enh_ack_generator.c
```diff
@@ -354,6 +355,14 @@ const uint8_t * nrf_802154_enh_ack_generator_create(const uint8_t * p_frame)
         frame_offsets.src_addr_size == EXTENDED_ADDRESS_SIZE,
         &ie_data_len);
 
+    //
```
jciupis: Empty comment

src/mac_features/nrf_802154_csl_phase_injector.c
```diff
@@ -38,36 +38,60 @@
...
+#define SYMBOLS_PER_SECOND       62500
+
+/**
+ * @brief Gets the time of the next receive delayed timeslot trigger time in symbols.
+ *
+ * @returns  Amount of symbols for the next receive delayed timeslot trigger time.
+ */
+static uint32_t nrf_802154_csl_phase_injector_phase_symbols_get(void)
+{
+    return nrf_802154_rsch_get_next_scheduled_receive_time() * 1000000 / SYMBOLS_PER_SECOND;
```
jciupis: This code assumes that it is the responsibility of the higher layer to ensure the delayed timeslot is actually scheduled and makes sense in the context of CSL. Please add a comment that explains it here.

src/nrf_802154_config.h
```diff
@@ -460,6 +460,38 @@ extern "C" {
...
+/**
+ * @def NRF_802154_STATS_COUNT_ENERGY_DETECTED_EVENTS
+ *
+ * Configures if energy detected events will be counted in receive mode.
+ * When this option is enabled additional interrupts on energy detected events will occur
+ * increasing power consumption. The events counter is stored in
+ * @ref nrf_802154_stat_counters_t::received_energy_events field and can be retrieved by
+ * a call to @ref nrf_802154_stats_get or @ref nrf_802154_stat_counters_get.
+ */
+#ifndef NRF_802154_STATS_COUNT_ENERGY_DETECTED_EVENTS
```
jciupis: I think something went wrong with the rebase, this code has already been merged to the target branch.

src/nrf_802154_const.h
mikaczma: Name IE_HEADER_LENGTH may be confusing with LENGTH field in IE_HEADER, better will be change to i.e. OCTETS (according to 802.15.4-2015 Std)

src/nrf_802154_const.h
```diff
@@ -78,10 +78,20 @@
 #define FRAME_VERSION_2              0x20                                         ///< Bits containing the frame version 0b10.
 #define FRAME_VERSION_3              0x30                                         ///< Bits containing the frame version 0b11.
 
-#define IE_HEADER_LENGTH_MASK        0x3f                                         ///< Mask of bits containing the length of an IE header content.
+#define IE_HEADER_LENGTH_OFFSET      0                                            ///< Byte containing the Length field in IE header content
+#define IE_HEADER_LENGTH_MASK        0x7f                                         ///< Mask of bits containing the length of an IE header content.
+#define IE_HEADER_TYPE_OFFSET_0      0                                            ///< Byte containing the lower bits of Type field of IE header
```
ankuns: If this is related to spec's Figure 7-21 field ElementID this should have `ELEMENTID` in identifier instead of `TYPE`. Note there is a Type=0 field tn Figure 7-21

src/mac_features/nrf_802154_sec_key_manager.h
```diff
@@ -0,0 +1,119 @@
...
+    nrf_802154_sec_key_manager_key_descriptor_t key_descriptor;
+} nrf_802154_sec_key_manager_key_id_lookup_descriptor_t;
+
+/**
+ * @brief   Update CSL Phase field in IE header
+ * IEEE std 802.15.4-2015
+ *
+ * @param[in] p_ie_csl_header Pointer to IE CSL header
+ */
+bool nrf_802154_csl_phase_injector_ie_csl_header_phase_update(uint8_t * p_ie_csl_header);
```
ankuns: This prototype should not go into `nrf_802154_sec_key_manager.h`

src/nrf_802154_const.h
```diff
@@ -78,10 +78,20 @@
 #define FRAME_VERSION_2              0x20                                         ///< Bits containing the frame version 0b10.
 #define FRAME_VERSION_3              0x30                                         ///< Bits containing the frame version 0b11.
 
-#define IE_HEADER_LENGTH_MASK        0x3f                                         ///< Mask of bits containing the length of an IE header content.
+#define IE_HEADER_LENGTH_OFFSET      0                                            ///< Byte containing the Length field in IE header content
```
ankuns: Reference to standard IEEE Std. 802.15.4-2014 Figure 7-21 in comment will make easier to track where these values come from. This Applies to all constants you introduced.

src/mac_features/nrf_802154_sec_key_manager.c
```diff
@@ -0,0 +1,176 @@
...
+#include <string.h>
+#include "nrf_802154_sec_key_manager.h"
+#include "../nrf_802154_const.h"
+
+nrf_802154_sec_key_manager_key_id_lookup_descriptor_t * mp_key_id_lookup_list = NULL;
+uint8_t m_mac_pan_id[PAN_ID_LENGTH];
+uint8_t m_mac_coord_extended_addr[EXTENDED_ADDR_LENGTH];
+uint8_t m_mac_coord_short_addr[SHORT_ADDR_LENGTH];
+
+bool nrf_802154_sec_key_manager_lookup_procedure(
```
ankuns: Missing prototype in header file. Missing doc about parameters.

src/mac_features/nrf_802154_sec_key_manager.c
```diff
@@ -0,0 +1,176 @@
...
+    uint8_t                                               * key_source,
+    uint8_t                                                 key_index,
+    sec_key_device_addr_mode_t                              device_addr_mode,
+    uint8_t                                               * device_pan_id,
+    uint8_t                                               * device_addr,
+    nrf_802154_sec_key_manager_key_id_lookup_descriptor_t * key_id_lookup_descriptor)
+{
+    if (mp_key_id_lookup_list != NULL)
+    {
+        for (int i = 0; i < (sizeof(mp_key_id_lookup_list) / sizeof(mp_key_id_lookup_list[0])); i++)
```
ankuns: Use `size_t` from `stddef.h` instead of plain `int`. Generally we never use plain `int` unless it is strictly required by third-party library API. We use types from stdint.h and `size_t` where involved with sizeof operator.

src/mac_features/nrf_802154_sec_key_manager.c
```diff
@@ -0,0 +1,176 @@
...
+                                    key_source_length) == 0)
+                        {
+                            key_id_lookup_descriptor = &mp_key_id_lookup_list[i];
+                            return true;
+                        }
+                    }
+                    break;
+
+                default:
+                    break;
```
ankuns: If getting into default case is illegal, then please `assert(false)` If it has some meaning then comment there why default does nothing.

src/mac_features/nrf_802154_sec_key_manager.c
```diff
@@ -0,0 +1,176 @@
...
+    const uint8_t                                         * p_frame,
+    uint8_t                                                 key_id_mode,
+    uint8_t                                               * key_source,
+    uint8_t                                                 key_index,
+    sec_key_device_addr_mode_t                              device_addr_mode,
+    uint8_t                                               * device_pan_id,
+    uint8_t                                               * device_addr,
+    nrf_802154_sec_key_manager_key_id_lookup_descriptor_t * key_id_lookup_descriptor)
+{
+    if (mp_key_id_lookup_list != NULL)
```
ankuns: Definitely please split this divine function into smaller parts. Use static functions as sub-procedures. Currently the complexity is unacceptably high.

src/mac_features/nrf_802154_sec_key_manager.c
```diff
@@ -0,0 +1,185 @@
...
+
+                    if ((device_addr_mode == NONE))
+                    {
+                        if (frame_type == FRAME_TYPE_BEACON)
+                        {
+                            memcpy(p_device_addr, m_mac_coord_extended_addr, EXTENDED_ADDR_LENGTH);
+                        }
+                        else
+                        {
+                            if ((m_mac_coord_short_addr[0] == 0xff) &&
```
ankuns: There should be api in PIB to get _macCoordShortAddress_ (see spec Table 8-81). You should use this api, provide if necessary.

src/mac_features/nrf_802154_sec_key_manager.c
```diff
@@ -0,0 +1,185 @@
...
+#define FRAME_COUNTER_LENGTH 4  ///< As defined in 802.15.4 Std - Chapter 9.4.2 & Table 9-10
+#define KEY_LENGTH           16 ///< As defined in 802.15.4 Std - Table 9-10
+#define EXTENDED_ADDR_LENGTH 8  ///< As defined in 802.15.4 Std - Chapter 7.1
+#define SHORT_ADDR_LENGTH    2  ///< As defined in 802.15.4 Std - Table 9-14
+#define PAN_ID_LENGTH        2  ///< As defined in 802.15.4 Std - Table 9-14
+
+static nrf_802154_sec_key_manager_key_id_lookup_descriptor_t * mp_key_id_lookup_list = NULL;
+static uint8_t m_mac_pan_id[PAN_ID_LENGTH];
+static uint8_t m_mac_coord_extended_addr[EXTENDED_ADDR_LENGTH];
+static uint8_t m_mac_coord_short_addr[SHORT_ADDR_LENGTH];
```
ankuns: This should be in the PIB module

src/mac_features/nrf_802154_sec_key_manager.c
```diff
@@ -0,0 +1,185 @@
...
+#include "../nrf_802154_const.h"
+
+#define FRAME_COUNTER_LENGTH 4  ///< As defined in 802.15.4 Std - Chapter 9.4.2 & Table 9-10
+#define KEY_LENGTH           16 ///< As defined in 802.15.4 Std - Table 9-10
+#define EXTENDED_ADDR_LENGTH 8  ///< As defined in 802.15.4 Std - Chapter 7.1
+#define SHORT_ADDR_LENGTH    2  ///< As defined in 802.15.4 Std - Table 9-14
+#define PAN_ID_LENGTH        2  ///< As defined in 802.15.4 Std - Table 9-14
+
+static nrf_802154_sec_key_manager_key_id_lookup_descriptor_t * mp_key_id_lookup_list = NULL;
+static uint8_t m_mac_pan_id[PAN_ID_LENGTH];
```
ankuns: There is `nrf_802154_pib_pan_id_get` in PIB already.

src/mac_features/nrf_802154_sec_key_manager.c
```diff
@@ -0,0 +1,185 @@
...
+static uint8_t m_mac_coord_extended_addr[EXTENDED_ADDR_LENGTH];
```
ankuns: This should be in the PIB module

src/mac_features/nrf_802154_sec_key_manager.c
```diff
@@ -0,0 +1,185 @@
...
+static uint8_t m_mac_coord_short_addr[SHORT_ADDR_LENGTH];
+
+bool nrf_802154_sec_key_manager_lookup_procedure(
+    const uint8_t                                         * p_frame,
+    uint8_t                                                 key_id_mode,
+    uint8_t                                               * p_key_source,
+    uint8_t                                                 key_index,
+    sec_key_device_addr_mode_t                              device_addr_mode,
+    uint8_t                                               * p_device_pan_id,
```
ankuns: if thiw was `const uint8_t * p_device_pan_id` compiler would detect flawed implementation and memcpy to this input parameter.

src/mac_features/nrf_802154_sec_key_manager.c
```diff
@@ -0,0 +1,185 @@
...
+                            case EXTENDED:
+                                addr_length = EXTENDED_ADDR_LENGTH;
+                                break;
+
+                            default:
+                                return false;
+                                break;
+                        }
+
+                        if (strncmp(p_device_addr, mp_key_id_lookup_list[i].key_device_address,
```
ankuns: Use `memcpy` instead of `strncmp`

src/mac_features/nrf_802154_sec_key_manager.c
```diff
@@ -40,39 +40,46 @@
...
-    uint8_t                                               * device_addr,
-    nrf_802154_sec_key_manager_key_id_lookup_descriptor_t * key_id_lookup_descriptor)
+    uint8_t                                               * p_device_pan_id,
+    uint8_t                                               * p_device_addr,
+    nrf_802154_sec_key_manager_key_id_lookup_descriptor_t * p_key_id_lookup_descriptor)
 {
     if (mp_key_id_lookup_list != NULL)
     {
-        for (int i = 0; i < (sizeof(mp_key_id_lookup_list) / sizeof(mp_key_id_lookup_list[0])); i++)
+        for (size_t i = 0; i < (sizeof(mp_key_id_lookup_list) / sizeof(mp_key_id_lookup_list[0]));
```
ankuns: `sizeof(mp_key_id_lookup_list) / sizeof(mp_key_id_lookup_list[0])` when we have 
```c
static nrf_802154_sec_key_manager_key_id_lookup_descriptor_t * mp_key_id_lookup_list
```
will surely be constant equal to 1. There is "number of entries" missing to indicate how many items should we iterate over.

src/mac_features/nrf_802154_sec_key_manager.c
```diff
@@ -163,23 +229,22 @@ bool nrf_802154_sec_key_manager_lookup_procedure(
                                 break;
                         }
 
-                        if (strncmp(p_key_source, mp_key_id_lookup_list[i].key_source,
-                                    key_source_length) == 0)
+                        if (memcmp(p_key_source, mp_key_id_lookup_list[i].key_source,
+                                   key_source_length) == 0)
                         {
-                            p_key_id_lookup_descriptor = &mp_key_id_lookup_list[i];
+                            p_key_descriptor = &(mp_key_id_lookup_list[i].key_descriptor);
```
ankuns: Taking into account other comments about way of returning key_descriptor, I think it should be
```c
*pp_key_descriptor = &(mp_key_id_lookup_list[i].key_descriptor);
```

src/mac_features/nrf_802154_sec_key_manager.c
```diff
@@ -163,23 +229,22 @@ bool nrf_802154_sec_key_manager_lookup_procedure(
...
-
-                default:
-                    assert(false);
-                    break;
+                }
+                break;
 
-            }
+            default:
+                assert(false);
```
ankuns: Th was my mistake advicing to assert. My fault, I'm sorry. Default case is covered by 9.2.2. d) so the procedure should simply return `false` (as spec states `FAILED`).

src/mac_features/nrf_802154_sec_key_manager.h
```diff
@@ -43,6 +43,12 @@
 extern "C" {
 #endif
 
+#define FRAME_COUNTER_LENGTH 4  ///< As defined in 802.15.4 Std - Chapter 9.4.2 & Table 9-10
+#define KEY_LENGTH           16 ///< As defined in 802.15.4 Std - Table 9-10
+#define EXTENDED_ADDR_LENGTH 8  ///< As defined in 802.15.4 Std - Chapter 7.1
+#define SHORT_ADDR_LENGTH    2  ///< As defined in 802.15.4 Std - Table 9-14
```
ankuns: We have `SHORT_ADDRESS_SIZE`, `EXTENDED_ADDRESS_SIZE`, `PAN_ID_SIZE` in `nrf_802154_const.h`. Please use them instead of those redundant definitions.

src/nrf_802154_encrypt.c
```diff
@@ -516,6 +519,51 @@ void nrf_802154_encrypt_aes_ccm_auth_transform_trigger(const uint8_t * p_frame)
...
 
+bool nrf_802154_encrypt_nonce_generate(const uint8_t * p_frame, bool is_tsch_mode,
+                                       uint8_t * p_nonce)
+{
+    if (p_nonce == NULL)
+    {
+        return false;
+    }
+
+    uint8_t nonce[NRF_802154_ENCRYPT_NONCE_SIZE];
```
ankuns: Is there any reason for not preparing nonce directly behind pointer `p_nonce`

src/nrf_802154_encrypt.c
```diff
@@ -516,6 +519,51 @@ void nrf_802154_encrypt_aes_ccm_auth_transform_trigger(const uint8_t * p_frame)
...
+            return false;
+        }
+
+        for (uint8_t i = 0; i < EXTENDED_ADDRESS_SIZE; i++)
+        {
+            nonce[i] = mhr_data.p_src_addr[EXTENDED_ADDRESS_SIZE - 1 - i];
+        }
+
+        const uint8_t * frame_counter = nrf_802154_frame_parser_frame_counter_get(p_frame);
+
```
ankuns: A comment referring to spec 9.3.1 and B.2 would make it easier to justify why you are reversing byte order for frame counter

src/nrf_802154_encrypt.c
```diff
@@ -516,6 +519,51 @@ void nrf_802154_encrypt_aes_ccm_auth_transform_trigger(const uint8_t * p_frame)
...
+        for (uint8_t i = 0; i < FRAME_COUNTER_SIZE; i++)
```
ankuns: `for (size_t i = 0; `
or
`for (uint_fast8_t i = 0;`

Could this loop be wrapped with some helper function
```c
void memcpy_rev(void * p_dst, const void * p_src, size_t n)
```

src/nrf_802154_encrypt.c
```diff
@@ -516,6 +519,51 @@ void nrf_802154_encrypt_aes_ccm_auth_transform_trigger(const uint8_t * p_frame)
...
+        {
+            nonce[i + EXTENDED_ADDRESS_SIZE] = frame_counter[FRAME_COUNTER_SIZE - 1 - i];
+        }
+
+        nonce[NRF_802154_ENCRYPT_NONCE_SIZE - 1] = nrf_802154_frame_parser_security_level_get(
```
ankuns: use 
```c
if (mhr_data.p_sec_ctrl == NULL)
{
   return false'
}
```
before. `nrf_802154_frame_parser_security_level_get` will dereference NULL poinetr without that.

src/nrf_802154_pib.h
```diff
@@ -171,6 +171,38 @@ const uint8_t * nrf_802154_pib_short_address_get(void);
  */
 void nrf_802154_pib_short_address_set(const uint8_t * p_short_address);
 
+/**
+ * @brief Gets the extended address of coordinator.
```
ankuns: Please add to doc original spelling `macCoordExtendedAddress` and reference to Table 8-81 in spec. Similar applies to other functions you've added to PIB. There are plenty of similar spelled fields in the pib, we use transformed names due to coding standard. Reference to original spelling would simplify catching intentions and look-ups between the code and the spec.

src/mac_features/nrf_802154_frame_parser.h
```diff
@@ -369,4 +369,34 @@ bool nrf_802154_frame_parser_is_csl_ie_header_available(const uint8_t * p_frame)
...
+ *
+ * @returns  Value of a given type of the @p p_frame.
+ *
+ */
+uint8_t nrf_802154_frame_parser_frame_type_get(const uint8_t * p_frame);
+
+/**
+ * @brief Gets the security level provided in frame security field
+ *
+ * @param[in] p_security_field  Pointer to a security field
```
ankuns: I think the spec uses the term "auxiliary security header" is used, please stick to terms used in spec as far as possible.

src/mac_features/nrf_802154_sec_key_manager.c
```diff
@@ -0,0 +1,246 @@
...
+                                case SHORT:
+                                    addr_length = SHORT_ADDRESS_SIZE;
+                                    break;
+
+                                case EXTENDED:
+                                    addr_length = EXTENDED_ADDRESS_SIZE;
+                                    break;
+
+                                default:
+                                    return false;
```
ankuns: Case `NONE` is very tricky. According to table 9-9 *secKeyDecideAddrMode* in PIB can have value NONE. What is the value of *secKeyDeviceAddress*, there is no address to compare with. On the other hand when `device_addr_mode` (aka specs 9.2.2 *DevideAddressingMode*) is NONE the `p_device_addr ` (*DeviceAddress*) was set to either *macCoordShortAddress* or *macCoordExtendedAddress*. 
@hubertmis please could you consult on this? I think that passing device_addr_mode == NONE is perfect, but we should not return false here. What to compare then, how long data etc?

src/nrf_802154_encrypt.c
```diff
@@ -527,37 +527,46 @@ bool nrf_802154_encrypt_nonce_generate(const uint8_t * p_frame, bool is_tsch_mod
         return false;
     }
 
-    uint8_t nonce[NRF_802154_ENCRYPT_NONCE_SIZE];
+    // uint8_t nonce[NRF_802154_ENCRYPT_NONCE_SIZE];
 
-    memset(nonce, 0, NRF_802154_ENCRYPT_NONCE_SIZE);
+    // memset(nonce, 0, NRF_802154_ENCRYPT_NONCE_SIZE);
```
ankuns: Remove commented-out code

src/nrf_802154_encrypt.c
```diff
@@ -527,37 +527,46 @@ bool nrf_802154_encrypt_nonce_generate(const uint8_t * p_frame, bool is_tsch_mod
...
 
+        p_nonce[0] = mhr_data.p_src_addr[EXTENDED_ADDRESS_SIZE - 1];
+        p_nonce[1] = mhr_data.p_src_addr[EXTENDED_ADDRESS_SIZE - 2];
+        p_nonce[2] = mhr_data.p_src_addr[EXTENDED_ADDRESS_SIZE - 3];
+        p_nonce[3] = mhr_data.p_src_addr[EXTENDED_ADDRESS_SIZE - 4];
+        p_nonce[4] = mhr_data.p_src_addr[EXTENDED_ADDRESS_SIZE - 5];
+        p_nonce[5] = mhr_data.p_src_addr[EXTENDED_ADDRESS_SIZE - 6];
+        p_nonce[6] = mhr_data.p_src_addr[EXTENDED_ADDRESS_SIZE - 7];
+        p_nonce[7] = mhr_data.p_src_addr[EXTENDED_ADDRESS_SIZE - 8];
+
```
ankuns: This is not what I intended. I asked for extracting copying with reverse order into separate function for example with prototype
```c
void memcpy_rev(void * p_dst, const void * p_src, size_t n)
```
And use this in two places for setting *Source Address* and *Frame Counter* fields in nonce.

src/nrf_802154_encrypt.c
```diff
@@ -527,37 +527,46 @@ bool nrf_802154_encrypt_nonce_generate(const uint8_t * p_frame, bool is_tsch_mod
...
 
-        for (uint8_t i = 0; i < FRAME_COUNTER_SIZE; i++)
-        {
-            nonce[i + EXTENDED_ADDRESS_SIZE] = frame_counter[FRAME_COUNTER_SIZE - 1 - i];
-        }
+        // Byte order for Frame Counter gets reversed as defined in 802.15.4-2015 Std Chapters 9.3.1 and Annex B.2
+        p_nonce[0 + EXTENDED_ADDRESS_SIZE] = frame_counter[FRAME_COUNTER_SIZE - 1];
+        p_nonce[1 + EXTENDED_ADDRESS_SIZE] = frame_counter[FRAME_COUNTER_SIZE - 2];
+        p_nonce[2 + EXTENDED_ADDRESS_SIZE] = frame_counter[FRAME_COUNTER_SIZE - 3];
+        p_nonce[3 + EXTENDED_ADDRESS_SIZE] = frame_counter[FRAME_COUNTER_SIZE - 4];
```
ankuns: The same comment as for above *Source Address*

PR #318, opened 2019-11-21
Implement antenna diversity module

JakubPegza:
* Antenna diversity module created and added to packsc.
  * ant_div option added to packsc
  * Antenna diversity implements:
    - Pinout configuration
    - Initialization
    - Antenna selection
    - Retrieving current antenna
  * Stubs provided for non-ant_div builds

On-desk manual tests passed.

### DIFF
--- a/src/nrf_802154_ant_div.c
+++ b/src/nrf_802154_ant_div.c
@@ -1,0 +1,111 @@
+/* Copyright (c) 2017 - 2019, Nordic Semiconductor ASA
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *   1. Redistributions of source code must retain the above copyright notice, this
+ *      list of conditions and the following disclaimer.
+ *
+ *   2. Redistributions in binary form must reproduce the above copyright notice,
+ *      this list of conditions and the following disclaimer in the documentation
+ *      and/or other materials provided with the distribution.
+ *
+ *   3. Neither the name of Nordic Semiconductor ASA nor the names of its
+ *      contributors may be used to endorse or promote products derived from
+ *      this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ */
+
+/**
+ * @file
+ *   This file implements the 802.15.4 antenna diversity module.
+ *
+ */
+#include <assert.h>
+
+#include "nrf_802154_ant_div.h"
+#include "nrf_error.h"
+#include "nrf_gpio.h"
+
+#ifdef ENABLE_ANT_DIV
+
+static nrf_802154_ant_div_config_t m_ant_div_config =      /**< Antenna Diversity configuration */
+{
+    .ant_sel_pin = NRF_ANT_DIV_ANT_SEL_DEFAULT_PIN
+};
+
+int32_t nrf_802154_ant_div_init()
+{
+    nrf_gpio_cfg_output(m_ant_div_config.ant_sel_pin);
+    return NRF_SUCCESS;
+}
+
+int32_t nrf_802154_ant_div_set_config(nrf_802154_ant_div_config_t *p_ant_div_config)
+{
+    assert(p_ant_div_config != NULL);
+    m_ant_div_config = *p_ant_div_config;
+    return NRF_SUCCESS;
+}
+
+int32_t nrf_802154_ant_div_get_config(nrf_802154_ant_div_config_t *p_ant_div_config)
+{
+    assert(p_ant_div_config != NULL);
+    *p_ant_div_config = m_ant_div_config;
+    return NRF_SUCCESS;
+}
+
+int32_t nrf_802154_ant_div_set_antenna(nrf_802154_ant_div_antenna_t antenna)
+{
+    nrf_gpio_pin_write(m_ant_div_config.ant_sel_pin, antenna);
+    return NRF_SUCCESS;
+}
+
+int32_t nrf_802154_ant_div_get_antenna(nrf_802154_ant_div_antenna_t *antenna)
+{
+    *antenna = (NRF_GPIO->OUT >> m_ant_div_config.ant_sel_pin) & 0x01;
+    return NRF_SUCCESS;
+}
+#else //ENABLE_ANT_DIV
+
+int32_t nrf_802154_ant_div_init()
+{
+    return NRF_ERROR_NOT_SUPPORTED;
+}
+
+int32_t nrf_802154_ant_div_set_config(nrf_802154_ant_div_config_t *p_ant_div_config)
+{
+    (void)p_ant_div_config;
+    return NRF_ERROR_NOT_SUPPORTED;
+}
+
+int32_t nrf_802154_ant_div_get_config(nrf_802154_ant_div_config_t *p_ant_div_config)
+{
+    (void)p_ant_div_config;
+    return NRF_ERROR_NOT_SUPPORTED;
+}
+
+int32_t nrf_802154_ant_div_set_antenna(nrf_802154_ant_div_antenna_t antenna)
+{
+    (void)antenna;
+    return NRF_ERROR_NOT_SUPPORTED;
+}
+
+int32_t nrf_802154_ant_div_get_antenna(nrf_802154_ant_div_antenna_t *antenna)
+{
+    (void)antenna;
+    return NRF_ERROR_NOT_SUPPORTED;
+}
+
+#endif //ENABLE_ANT_DIV
--- a/src/nrf_802154_ant_div.c
+++ b/src/nrf_802154_ant_div.c
@@ -1,4 +1,4 @@
-/* Copyright (c) 2017 - 2019, Nordic Semiconductor ASA
+/* Copyright (c) 2019, Nordic Semiconductor ASA
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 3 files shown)


src/nrf_802154_ant_div.h
```diff
@@ -0,0 +1,101 @@
+/* Copyright (c) 2017, Nordic Semiconductor ASA
```
jciupis: Please update the license

src/nrf_802154_ant_div.h
```diff
@@ -0,0 +1,101 @@
...
+ */
+#include "stdint.h"
+
+#define NRF_ANT_DIV_ANT_SEL_DEFAULT_PIN 23
+
+/**
+ * @brief Configuration structure for antenna diversity
+ */
+typedef struct{
+    uint8_t ant_sel_pin;        /* Pin used for antenna selection */
```
jciupis: Do we expect this structure to grow as the implementation proceeds?
* If the answer is positive, please have a look how the FEM module stores configuration options for the pins it uses and for its interface itself and get inspired ;)
* If the answer is no, simple `typedef uint8_t nrf_802154_and_div_config_t` is enough.

src/nrf_802154_ant_div.h
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * @brief Initializes antenna diversity module.
+ *
+ * If pinout other than default is to be used, @ref nrf_802154_ant_div_set_config 
+ * should be called before this function.
+ *  
+ * @retval ::NRF_SUCCESS             Antenna switched successfully
+ * @retval ::NRF_ERROR_NOT_SUPPORTED Antenna diversity module is not supported
+ */
+int32_t nrf_802154_ant_div_init();
```
jciupis: If you're planning to return non-negative values only, `uint32_t` is the better choice as it is unambiguous then. Otherwise, if returning `int32_t` is necessary to conform to MPSL API or its coding standards, I think you might consider returning `-::NRF_ERROR_NOT_SUPPORTED`. This comment refers to all functions in this module.

src/nrf_802154_ant_div.h
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * @brief Get currently used antenna.
+ *
+ * @param[out] antenna         Currently used antenna. Not modified if antenna diversity is not supported.
+ * 
+ * @retval ::NRF_SUCCESS             Antenna switched successfully
+ * @retval ::NRF_ERROR_NOT_SUPPORTED Antenna diversity module is not supported
+ */
+int32_t nrf_802154_ant_div_get_antenna(nrf_802154_ant_div_antenna_t *antenna);
```
jciupis: Pointers should be prefixed with `p_`

src/nrf_802154_ant_div.h
```diff
@@ -0,0 +1,101 @@
...
+ * @retval ::NRF_ERROR_NOT_SUPPORTED Antenna diversity module is not supported
+ */
+int32_t nrf_802154_ant_div_set_antenna(nrf_802154_ant_div_antenna_t antenna);
+
+/**
+ * @brief Get currently used antenna.
+ *
+ * @param[out] antenna         Currently used antenna. Not modified if antenna diversity is not supported.
+ * 
+ * @retval ::NRF_SUCCESS             Antenna switched successfully
```
jciupis: Comment copy&paste mistake here

src/nrf_802154_ant_div.h
jciupis: Please add an empty line at the end of the file

src/nrf_802154_ant_div.c
```diff
@@ -0,0 +1,111 @@
+/* Copyright (c) 2017 - 2019, Nordic Semiconductor ASA
```
jciupis: I think you might leave just `2019` here

src/nrf_802154_ant_div.h
```diff
@@ -0,0 +1,101 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ */
+
+/**
+ * @brief 802.15.4 antenna diversity module.
+ */
+#include "stdint.h"
+
+#define NRF_ANT_DIV_ANT_SEL_DEFAULT_PIN 23
```
jciupis: ANT is actually `a popular proprietary ultra-low power wireless solution operating in the 2.4GHz ISM`, citing [nordicsemi.no.](https://www.nordicsemi.com/Products/Low-power-short-range-wireless/ANT) It might be therefore a bit confusing what this macro and the enum values refer to. I'd change it to `NRF_802154_ANT_DIV_DEFAULT_ANT_SEL_PIN`

src/nrf_802154_ant_div.c
```diff
@@ -0,0 +1,111 @@
...
+    nrf_gpio_pin_write(m_ant_div_config.ant_sel_pin, antenna);
+    return NRF_SUCCESS;
+}
+
+int32_t nrf_802154_ant_div_get_antenna(nrf_802154_ant_div_antenna_t *antenna)
+{
+    *antenna = (NRF_GPIO->OUT >> m_ant_div_config.ant_sel_pin) & 0x01;
+    return NRF_SUCCESS;
+}
+#else //ENABLE_ANT_DIV
```
jciupis: I'd move this part to the header file as the FEM module does.

src/nrf_802154_ant_div.c
```diff
@@ -0,0 +1,111 @@
...
+    return NRF_ERROR_NOT_SUPPORTED;
+}
+
+int32_t nrf_802154_ant_div_get_antenna(nrf_802154_ant_div_antenna_t *antenna)
+{
+    (void)antenna;
+    return NRF_ERROR_NOT_SUPPORTED;
+}
+
+#endif //ENABLE_ANT_DIV
```
jciupis: Please add an empty line at the end of the file

src/nrf_802154_ant_div.c
```diff
@@ -0,0 +1,111 @@
...
+int32_t nrf_802154_ant_div_get_config(nrf_802154_ant_div_config_t *p_ant_div_config)
+{
+    assert(p_ant_div_config != NULL);
+    *p_ant_div_config = m_ant_div_config;
+    return NRF_SUCCESS;
+}
+
+int32_t nrf_802154_ant_div_set_antenna(nrf_802154_ant_div_antenna_t antenna)
+{
+    nrf_gpio_pin_write(m_ant_div_config.ant_sel_pin, antenna);
```
jciupis: Possibly worth checking if the provided argument actually falls into the expected range (i.e 0-1) before writing it to GPIO

src/nrf_802154_ant_div.c
```diff
@@ -0,0 +1,111 @@
...
+    return NRF_SUCCESS;
+}
+
+int32_t nrf_802154_ant_div_get_antenna(nrf_802154_ant_div_antenna_t *antenna)
+{
+    *antenna = (NRF_GPIO->OUT >> m_ant_div_config.ant_sel_pin) & 0x01;
```
jciupis: Why not use the GPIO HAL?
```
*antenna = (nrf_802154_ant_div_antenna_t)nrf_gpio_pin_read(m_ant_div_congih.ant_sel_pin);
```

src/nrf_802154_ant_div.h
JakubPegza: The structure will probably grow, although, unlike FEM, only 1 pin should be used in this module. Enable and polarity configuration will not be necessary for this pin as well. Considering that, I think there's no need for separate structure for pin configuration.

src/nrf_802154_ant_div.h
```diff
@@ -0,0 +1,101 @@
...
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ */
+
+/**
+ * @brief 802.15.4 antenna diversity module.
+ */
+#include "stdint.h"
```
ankuns: `#include <stdint.h>`

src/nrf_802154_ant_div.h
ankuns: Please make pinology configurable for target board by using
```c
#ifndef NRF_802154_ANT_DIV_DEFAULT_ANT_SEL_PIN
#define NRF_802154_ANT_DIV_DEFAULT_ANT_SEL_PIN 23
#endif
```

src/nrf_802154_ant_div.h
ankuns: Please use prototype with explicit `void` if no arguments (`nrf_802154_ant_div_init(void)`). In C this has different meaning than in C++.

src/nrf_802154_ant_div.h
```diff
@@ -0,0 +1,101 @@
...
+ * @brief Sets the antenna diversity configuration.
+ * 
+ * Should not be called after @ref nrf_802154_ant_div_init.
+ * 
+ * @param[in] ant_div_config   Pointer to the antenna diversity configuration structure.
+ *
+ * @retval ::NRF_SUCCESS             Configuration set successfully
+ * @retval ::NRF_ERROR_NOT_SUPPORTED Antenna diversity module is not supported 
+ */
+int32_t nrf_802154_ant_div_set_config(nrf_802154_ant_div_config_t *p_ant_div_config);
```
ankuns: Please use **const** for pointer parameters that are not changing in function call. This will save a lot of RAM as the pin configuration may be stored in flash. `const nrf_802154_ant_div_config_t * p_ant_div_config`.
Please put space between `*` `and `p_ant_div_config` (FCS 3.1.5)

Please consider renaming into nrf_802154_ant_div_config_set (verb at the end, FCS 6.1.10)

src/nrf_802154_ant_div.h
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * @brief Retrieves te antenna diversity configuration.
+ *
+ * @param[out] ant_div_config  Pointer to configuration structure to be populated.
+ * 
+ * @retval ::NRF_SUCCESS             Configuration retrieved successfully
+ * @retval ::NRF_ERROR_NOT_SUPPORTED Antenna diversity module is not supported
+ */
+int32_t nrf_802154_ant_div_get_config(nrf_802154_ant_div_config_t *p_ant_div_config);
```
ankuns: Please consider renaming into nrf_802154_ant_div_config_get (verb at the end, FCS 6.1.10)

src/nrf_802154_ant_div.c
```diff
@@ -0,0 +1,90 @@
...
+    assert(p_ant_div_config != NULL);
+    *p_ant_div_config = m_ant_div_config;
+    return NRF_SUCCESS;
+}
+
+uint32_t nrf_802154_ant_div_antenna_set(nrf_802154_ant_div_antenna_t antenna)
+{
+    uint32_t status = NRF_SUCCESS;
+
+    if (NRF_ANT_DIV_ANTENNA_1 == antenna || NRF_ANT_DIV_ANTENNA_2 == antenna )
```
ankuns: Please use parentheses 
```c
 if ((NRF_ANT_DIV_ANTENNA_1 == antenna) || (NRF_ANT_DIV_ANTENNA_2 == antenna))
```

src/nrf_802154_ant_div.h
jciupis: If `ANT_SEL` is the only pin that is going to be configures by this module and it doesn't need any additional configuration fields then it sounds pretty much like the second case. Do you see the need to add any fields to this structure in the nearest future?

src/nrf_802154_ant_div.h
```diff
@@ -0,0 +1,149 @@
...
+
+/**
+ * @brief Configuration structure for antenna diversity
+ */
+typedef struct
+{
+    uint8_t ant_sel_pin; /* Pin used for antenna selection */
+} nrf_802154_ant_div_config_t;
+
+typedef enum
```
jciupis: Please add a brief documentation comment here.

src/nrf_802154_ant_div.h
JakubPegza: Most probably yes, for example GPIOTE channel.

jciupis: Alright then.

PR #319, opened 2019-11-22
Pr/log rx stats

ankuns:
Implemented stat counters for:
- number of sync events (number of times energy detected during receive)
- number of received preambles
- number of received frames with CRCOK and filtering passing

This implementation requires additional interrupts to be enabled during receive. Because this can increase power consumption, this counters may be disabled at compile time. Agreed with @hubertmis that for counters disabled at compile time corresponding fields in `nrf_802154_stat_counters_t` remain (but won't be incremented because of compile-time configuration). Justification is to have coherent public api struct `nrf_802154_stat_counters_t` if the driver is compiled into lib and then linked with application. The api used by app will be then independent of compile-time configuration used when radio driver lib was compiled.

### DIFF
--- a/src/nrf_802154_core.c
+++ b/src/nrf_802154_core.c
@@ -741,6 +741,9 @@ static nrf_802154_trx_receive_notifications_t make_trx_frame_receive_notificatio
 #if (NRF_802154_STATS_COUNT_ENERGY_DETECTED_EVENTS)
     result |= TRX_RECEIVE_NOTIFICATION_PRESTARTED;
 #endif
+#if (NRF_802154_STATS_COUNT_RECEIVED_PREAMBLES)
+    result |= TRX_RECEIVE_NOTIFICATION_STARTED;
+#endif
 
     if (nrf_802154_wifi_coex_is_enabled())
     {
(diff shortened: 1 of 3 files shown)


src/nrf_802154_core.c
```diff
@@ -1246,10 +1261,24 @@ void nrf_802154_trx_receive_frame_started(void)
...
+    nrf_802154_stat_counter_increment(received_preambles);
+#endif
 
-    /* Request boosted preconditions */
-    nrf_802154_rsch_crit_sect_prio_request(RSCH_PRIO_RX);
+    switch (nrf_802154_pib_coex_rx_request_mode_get())
+    {
+        case NRF_802154_COEX_RX_REQUEST_MODE_ENERGY_DETECTION:
+            nrf_802154_timer_sched_remove(&m_rx_prestarted_timer, NULL);
+        /* no break */
```
jciupis: The comment should probably be indented like the code

ankuns: I agree with you, it was written such way originally, but uncrustify does not like it.

PR #320, opened 2019-11-25
Eliminate various races and increase stability

jciupis:
This PR introduces a number of changes that eliminate the impact of various races that have affected the driver's stability negatively.

### DIFF
--- a/src/nrf_802154_critical_section.c
+++ b/src/nrf_802154_critical_section.c
@@ -111,13 +111,16 @@ static bool nested_critical_section_is_allowed_in_this_context(void)
 
 static bool critical_section_enter(bool forced)
 {
-    bool    result = false;
-    uint8_t cnt;
+    bool     result = false;
+    uint8_t  cnt;
+    uint32_t interrupt_state = __get_PRIMASK();
 
     if (forced ||
         (m_nested_critical_section_counter == 0) ||
         nested_critical_section_is_allowed_in_this_context())
     {
+        __disable_irq();
+
         do
         {
             cnt = __LDREXB(&m_nested_critical_section_counter);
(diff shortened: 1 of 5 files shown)


src/nrf_802154_critical_section.c
```diff
@@ -118,6 +119,10 @@ static bool critical_section_enter(bool forced)
         (m_nested_critical_section_counter == 0) ||
         nested_critical_section_is_allowed_in_this_context())
     {
+        nrf_802154_mcu_critical_state_t mcu_cs;
+
+        nrf_802154_mcu_critical_enter(mcu_cs);
```
ankuns: What do you think about moving `mcu_critital_enter`/`exit` outside of `if`. In current code `m_nested_critical_section_counter ` tested for == 0 may change before we get into increment code.

src/rsch/raal/softdevice/nrf_raal_softdevice.c
```diff
@@ -346,6 +350,45 @@ static inline bool timeslot_state_is(timeslot_state_t state)
...
 
+/**@brief Atomically check if timeslot can be requested safely. */
+static bool timeslot_can_be_requested(void)
+{
+    bool                            result = false;
+    nrf_802154_mcu_critical_state_t mcu_cs;
+
+    nrf_802154_mcu_critical_enter(mcu_cs);
+
+    do
```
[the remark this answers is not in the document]
jciupis: True, I guess that's the price you pay for fixing bugs iteratively

src/nrf_802154_critical_section.c
jciupis: I could change that, but it's unnecessary. If higher priority code interrupts this function between the `m_nested_critical_section_counter == 0` check and disabling interrupts, we can assume that it's going to enter the critical section and exit it without affecting the lower priority code. This way we also avoid running `nested_critical_section_is_allowed_in_this_context()` with disabled interrupts and disabling interrupts when critical section cannot be entered anyway.

PR #325, opened 2019-12-03
Pr/tx timestamps

ankuns:
Implemented time stamping of various events during transmit operation.

### DIFF
--- a/src/mac_features/nrf_802154_csma_ca.c
+++ b/src/mac_features/nrf_802154_csma_ca.c
@@ -263,9 +263,11 @@ void nrf_802154_csma_ca_start(const uint8_t * p_data)
 {
     nrf_802154_log_function_enter(NRF_802154_LOG_VERBOSITY_LOW);
 
+#if (NRF_802154_FRAME_TIMESTAMP_ENABLED)
     uint32_t ts = nrf_802154_timer_sched_time_get();
 
     nrf_802154_stat_timestamp_write(last_csmaca_start_timestamp, ts);
+#endif
 
     assert(!procedure_is_running());
 
--- a/src/mac_features/nrf_802154_csma_ca.c
+++ b/src/mac_features/nrf_802154_csma_ca.c
@@ -50,6 +50,7 @@
 #include "nrf_802154_pib.h"
 #include "nrf_802154_procedures_duration.h"
 #include "nrf_802154_request.h"
+#include "nrf_802154_stats.h"
 #include "mac_features/nrf_802154_frame_parser.h"
 #include "platform/random/nrf_802154_random.h"
 #include "rsch/nrf_802154_rsch.h"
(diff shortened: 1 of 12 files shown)


src/nrf_802154_stats.c
```diff
@@ -66,6 +66,11 @@ void nrf_802154_stat_counters_subtract(const nrf_802154_stat_counters_t * stat_c
     }
 }
 
+void nrf_802154_stat_timestamps_get(nrf_802154_stat_timestamps_t * stat_timestamps)
```
jciupis: `p_stat_timestamps`

src/nrf_802154.h
```diff
@@ -1410,6 +1410,13 @@ void nrf_802154_stat_counters_get(nrf_802154_stat_counters_t * stat_counters);
  */
 void nrf_802154_stat_counters_subtract(const nrf_802154_stat_counters_t * stat_counters);
 
+/**
+ * @brief Get time stamps of events gathered by the last operation.
+ *
+ * @param[out] stat_timestamps Structure that will be filled with current time stamps of events.
+ */
+void nrf_802154_stat_timestamps_get(nrf_802154_stat_timestamps_t * stat_timestamps);
```
jciupis: `p_stat_timestamps`

src/mac_features/nrf_802154_csma_ca.c
```diff
@@ -262,6 +263,10 @@ void nrf_802154_csma_ca_start(const uint8_t * p_data)
 {
     nrf_802154_log_function_enter(NRF_802154_LOG_VERBOSITY_LOW);
 
+    uint32_t ts = nrf_802154_timer_sched_time_get();
+
+    nrf_802154_stat_timestamp_write(last_csmaca_start_timestamp, ts);
```
jciupis: Is there any particular reason for not collapsing these two lines into a single one? Would lint complain about such syntax?
`nrf_802154_stat_timestamp_write(last_csmaca_start_timestamp, nrf_802154_timer_sched_time_get());`

src/nrf_802154_core.c
```diff
@@ -1776,15 +1861,35 @@ void nrf_802154_trx_standalone_cca_finished(bool channel_was_idle)
     nrf_802154_log_function_exit(NRF_802154_LOG_VERBOSITY_LOW);
 }
 
+void nrf_802154_trx_transmit_frame_ccastarted(void)
+{
+    nrf_802154_log_function_enter(NRF_802154_LOG_VERBOSITY_LOW);
+
+    assert(false);
```
jciupis: Why is this function implemented this way? Is `TRX_TRANSMIT_NOTIFICATION_CCASTARTED` currently unsupported/unused?

ankuns: I added notification ccastarted to trx when I thought it will be necessary. Currently it is not necessary, but I would like to leave this feature in trx for future. What do you recommend? __WEAK in trx? empty function here?

jciupis: It can stay this way, but please add a comment that explains it is unused

src/mac_features/nrf_802154_csma_ca.c
ankuns: Currently nrf_802154_stat_timestamp_write is a macro. If we ever needed some enter/exit critical in the macro there we would have problem as timer_sched_time_get can be time consuming. With current implementation it would indeed looked better. Do you really want the change?

jciupis: I'm OK to leave it as it is

ankuns:
@jciupis Regarding to your general comment, NRF_802154_FRAME_TIMESTAMP_ENABLED will be in next commit

PR #326, opened 2019-12-03
Fix problem when CPU is halted during frame RX and FEM is disabled

hubertmis:
If CPU is halted when a frame is received, the timer responsible for
ACK transmission may overflow. Overflow of the timer was not detected
in software what lead to the driver hanging in the TX_ACK state.

Detection of timer overflow has been added to the code of the driver.


hubertmis:
Regression test

hubertmis:
Regression test

hubertmis:
The build failed due to build server problem, but stress test is passing now. Merging

PR #327, opened 2019-12-05
Feature/ant div pib

JakubPegza:
Reworked and expanded antenna diversity module:
 * PIB configuration added
 * Automatic antenna updates by TRX implemented
 * Reworked previous ant_div API
 * Reworked conditional compilation
 * Added public antenna diversity API

### DIFF
--- a/nRF-IEEE-802.15.4-radio-driver.packsc
+++ b/nRF-IEEE-802.15.4-radio-driver.packsc
@@ -7,7 +7,7 @@
         ],
         "_headers": [
             "src/nrf_802154.h",
-            "src/nrf_802154_ant_div.h",
+            "src/nrf_802154_ant_diversity.h",
             "src/nrf_802154_core_hooks.h",
             "src/nrf_802154_critical_section.h",
             "src/nrf_802154_debug.h",
@@ -46,7 +46,7 @@
                 ],
                 "_files": [
                     "src/nrf_802154.c",
-                    "src/nrf_802154_ant_div.c",
+                    "src/nrf_802154_ant_diversity.c",
                     "src/nrf_802154_core.c",
                     "src/nrf_802154_core_hooks.c",
                     "src/nrf_802154_critical_section.c",
@@ -98,7 +98,7 @@
                 ],
                 "_files": [
                     "src/nrf_802154.c",
-                    "src/nrf_802154_ant_div.c",
+                    "src/nrf_802154_ant_diversity.c",
                     "src/nrf_802154_core.c",
                     "src/nrf_802154_core_hooks.c",
                     "src/nrf_802154_critical_section.c",
@@ -147,7 +147,7 @@
                 ],
                 "_files": [
                     "src/nrf_802154.c",
-                    "src/nrf_802154_ant_div.c",
+                    "src/nrf_802154_ant_diversity.c",
                     "src/nrf_802154_core.c",
                     "src/nrf_802154_core_hooks.c",
                     "src/nrf_802154_critical_section.c",
@@ -467,11 +467,11 @@
         ]
     },
 
-    "ant_div": {
+    "ant_diversity": {
         "_class": "nRF_drv_radio_802_15_4",
         "_description": "This option enables antenna diversity.",
         "_defines": [
-            "ENABLE_ANT_DIV"
+            "ENABLE_ANT_DIVERSITY"
         ]
     },
 
--- a/src/mac_features/nrf_802154_delayed_trx.c
+++ b/src/mac_features/nrf_802154_delayed_trx.c
@@ -368,7 +368,6 @@ static void tx_timeslot_started_callback(rsch_dly_ts_id_t dly_ts_id)
 static void rx_timeslot_started_callback(rsch_dly_ts_id_t dly_ts_id)
 {
     nrf_802154_log_function_enter(NRF_802154_LOG_VERBOSITY_HIGH);
-    bool update_result;
 
     assert(dly_ts_id == RSCH_DLY_RX);
 
@@ -378,14 +377,7 @@ static void rx_timeslot_started_callback(rsch_dly_ts_id_t dly_ts_id)
         {
             nrf_802154_pib_channel_set(m_rx_channel);
 
-            #if ENABLE_ANT_DIV
-            update_result = nrf_802154_request_channel_update() &&
-                            nrf_802154_request_antenna_update();
-            #else // ENABLE_ANT_DIV
-            update_result = nrf_802154_request_channel_update();
-            #endif  // ENABLE_ANT_DIV
-
-            if (update_result)
+            if (nrf_802154_request_channel_update())
             {
                 (void)nrf_802154_request_receive(NRF_802154_TERM_802154,
                                                  REQ_ORIG_DELAYED_TRX,
--- a/src/mac_features/nrf_802154_delayed_trx.c
+++ b/src/mac_features/nrf_802154_delayed_trx.c
@@ -377,12 +377,13 @@ static void rx_timeslot_started_callback(rsch_dly_ts_id_t dly_ts_id)
         case DELAYED_TRX_OP_STATE_PENDING:
         {
             nrf_802154_pib_channel_set(m_rx_channel);
-            
+
             #if ENABLE_ANT_DIV
-            update_result = nrf_802154_request_channel_update() && nrf_802154_request_antenna_update();
+            update_result = nrf_802154_request_channel_update() &&
+                            nrf_802154_request_antenna_update();
             #else // ENABLE_ANT_DIV
             update_result = nrf_802154_request_channel_update();
-            #endif // ENABLE_ANT_DIV
+            #endif  // ENABLE_ANT_DIV
 
             if (update_result)
             {
--- a/src/mac_features/nrf_802154_delayed_trx.c
+++ b/src/mac_features/nrf_802154_delayed_trx.c
@@ -368,6 +368,7 @@ static void tx_timeslot_started_callback(rsch_dly_ts_id_t dly_ts_id)
 static void rx_timeslot_started_callback(rsch_dly_ts_id_t dly_ts_id)
 {
     nrf_802154_log_function_enter(NRF_802154_LOG_VERBOSITY_HIGH);
+    bool update_result;
 
     assert(dly_ts_id == RSCH_DLY_RX);
 
@@ -376,8 +377,14 @@ static void rx_timeslot_started_callback(rsch_dly_ts_id_t dly_ts_id)
         case DELAYED_TRX_OP_STATE_PENDING:
         {
             nrf_802154_pib_channel_set(m_rx_channel);
+            
+            #if ENABLE_ANT_DIV
+            update_result = nrf_802154_request_channel_update() && nrf_802154_request_antenna_update();
+            #else // ENABLE_ANT_DIV
+            update_result = nrf_802154_request_channel_update();
+            #endif // ENABLE_ANT_DIV
 
-            if (nrf_802154_request_channel_update())
+            if (update_result)
             {
                 (void)nrf_802154_request_receive(NRF_802154_TERM_802154,
                                                  REQ_ORIG_DELAYED_TRX,
--- a/src/nrf_802154.c
+++ b/src/nrf_802154.c
@@ -308,58 +308,31 @@ uint32_t nrf_802154_ant_div_mode_set(nrf_802154_ant_div_mode_t mode)
     return nrf_802154_pib_ant_div_mode_set(mode);
 }
 
-uint32_t nrf_802154_ant_div_mode_get(nrf_802154_ant_div_mode_t *p_mode)
+uint32_t nrf_802154_ant_div_mode_get(nrf_802154_ant_div_mode_t * p_mode)
 {
     return nrf_802154_pib_ant_div_mode_get(p_mode);
 }
 
 uint32_t nrf_802154_antenna_set(nrf_802154_ant_div_antenna_t antenna)
 {
-    nrf_802154_ant_div_mode_t mode;
-    uint32_t result;
-    
-    result = nrf_802154_pib_ant_div_mode_get(&mode);
-    
-    // Check if antenna diversity is supported
-    if (result == NRF_SUCCESS)
-    {
-        // Check if antenna diversity is in manual mode
-        if (mode != NRF_ANT_DIV_MODE_MANUAL)
-        {
-            result = NRF_ERROR_INVALID_STATE;
-        }
-        else
-        {
-            result = nr_802154_ant_div_antenna_set(antenna);
-        }
-    }
-
-    return result;
+    return nrf_802154_pib_ant_div_antenna_set(antenna);
 }
 
 uint32_t nrf_802154_antenna_get(nrf_802154_ant_div_antenna_t * p_antenna)
 {
-    return nrf_802154_ant_div_antenna_get(p_antenna);
+    return nrf_802154_pib_ant_div_antenna_get(p_antenna);
 }
 
-uint32_t nrf_802154_ant_div_cfg_set(const nrf_802154_ant_div_config_t *p_config)
+uint32_t nrf_802154_ant_div_cfg_set(const nrf_802154_ant_div_config_t * p_config)
 {
     return nrf_802154_ant_div_config_set(p_config);
 }
 
-uint32_t nrf_802154_ant_div_cfg_get(nrf_802154_ant_div_config_t *p_config)
+uint32_t nrf_802154_ant_div_cfg_get(nrf_802154_ant_div_config_t * p_config)
 {
     return nrf_802154_ant_div_config_get(p_config);
 }
 
-uint32_t nrf_802154_ant_div_enable(bool enabled)
-{
-    return nrf_802154_pib_ant_div_enable_set(enabled);
-}
-uint32_t nrf_802154_ant_div_is_enabled(bool *p_enabled)
-{
-    return nrf_802154_pib_ant_div_enable_get(p_enabled);
-}
 #endif // ENABLE_ANT_DIV
 
 nrf_802154_state_t nrf_802154_state_get(void)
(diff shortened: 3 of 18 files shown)


src/mac_features/nrf_802154_delayed_trx.c
```diff
@@ -377,7 +378,14 @@ static void rx_timeslot_started_callback(rsch_dly_ts_id_t dly_ts_id)
         {
             nrf_802154_pib_channel_set(m_rx_channel);
 
-            if (nrf_802154_request_channel_update())
+            #if ENABLE_ANT_DIV
```
czeslawmakarski: You can remove indentation from lines with preprocessor directives to make it more visible.

nRF-IEEE-802.15.4-radio-driver.packsc
```diff
@@ -463,6 +467,14 @@
         ]
     },
 
+    "ant_div": {
```
czeslawmakarski: Consider changing `div` to `diversity` so nobody confuses it with `division`.

src/nrf_802154.c
```diff
@@ -301,6 +301,52 @@ void nrf_802154_fem_control_cfg_get(nrf_802154_fem_control_cfg_t * p_cfg)
...
+nrf_802154_ant_div_mode_t nrf_802154_ant_div_mode_get(void)
+{
+    return nrf_802154_pib_ant_div_mode_get();
+}
+
+bool nrf_802154_antenna_set(nrf_802154_ant_div_antenna_t antenna)
+{
+    bool result = nrf_802154_pib_ant_div_antenna_set(antenna);
+
+    if (result && NRF_802154_ANT_DIV_MODE_MANUAL == nrf_802154_pib_ant_div_mode_get())
```
czeslawmakarski: `if ((result) && (NRF_802154_ANT_DIV_MODE_MANUAL == nrf_802154_pib_ant_div_mode_get()))`. I think `prettify.sh` would catch it anyway.

src/nrf_802154_ant_div.c
```diff
@@ -0,0 +1,80 @@
...
+ * @file
+ *   This file implements the 802.15.4 antenna diversity module.
+ *
+ */
+#include <assert.h>
+
+#include "nrf_802154_ant_div.h"
+#include "nrf_gpio.h"
+
+static nrf_802154_ant_div_config_t m_ant_div_config = /**< Antenna Diversity configuration */
```
czeslawmakarski: Consider placing initialisation in the `_init()` function, to preserve flash memory (initialised variables go to `.data` and so the whole structure shall be placed there, not only `.ant_sel_pin` field.

src/nrf_802154_core.c
```diff
@@ -2254,3 +2254,28 @@ bool nrf_802154_core_last_rssi_measurement_get(int8_t * p_rssi)
...
+
+        nrf_802154_critical_section_exit();
+    }
+
+    nrf_802154_log_function_exit(NRF_802154_LOG_VERBOSITY_LOW);
+
+    return result;
+}
+
+#endif // ENABLE_ANT_DIV
```
czeslawmakarski: No newline at the end of line, Keil wouldn't like it. ;)

src/nrf_802154_request_swi.c
```diff
@@ -77,6 +77,9 @@ typedef enum
     REQ_TYPE_CONTINUOUS_CARRIER,
     REQ_TYPE_MODULATED_CARRIER,
     REQ_TYPE_BUFFER_FREE,
+    #if ENABLE_ANT_DIV
```
czeslawmakarski: You can remove the indentation so it is more visible.

src/nrf_802154_request_swi.c
```diff
@@ -148,6 +151,13 @@ typedef struct
             bool    * p_result; ///< Buffer free request result.
         } buffer_free;          ///< Buffer free request details.
 
+        #if ENABLE_ANT_DIV
```
czeslawmakarski: Ditto.

src/nrf_802154_request_swi.c
```diff
@@ -662,6 +699,12 @@ static void irq_handler_req_event(void)
                     nrf_802154_core_notify_buffer_free(p_slot->data.buffer_free.p_data);
                 break;
 
+            #if ENABLE_ANT_DIV
```
czeslawmakarski: Ditto.

src/nrf_802154_types.h
```diff
@@ -205,6 +205,34 @@ typedef uint8_t nrf_802154_ifs_mode_t;
...
+ *
+ * Possible values:
+ * - @ref NRF_802154_ANT_DIV_MODE_DISABLED,
+ * - @ref NRF_802154_ANT_DIV_MODE_ANTENNA_1,
+ * - @ref NRF_802154_ANT_DIV_MODE_ANTENNA_2,
+ * - @ref NRF_802154_ANT_DIV_MODE_MANUAL
+ */
+typedef uint8_t nrf_802154_ant_div_mode_t;
+
+#define NRF_802154_ANT_DIV_MODE_DISABLED  0x00 // !< Antenna diversity is disabled
```
ankuns: ... and which antenna is to be used then?

src/nrf_802154_trx.c
```diff
@@ -695,6 +698,42 @@ void nrf_802154_trx_disable(void)
     nrf_802154_log_function_exit(NRF_802154_LOG_VERBOSITY_LOW);
 }
 
+#if ENABLE_ANT_DIV
+void nrf_802154_trx_antenna_update(void)
+{
+    bool                      result = true;
```
ankuns: Is this operation allowed when `m_trx_state == TRX_STATE_DISABLED`? In this state we usually have no timeslot and other drivers control radio, fem, ... and possibly antenna diversity (through fem).

src/nrf_802154_trx.c
```diff
@@ -900,6 +939,11 @@ void nrf_802154_trx_receive_frame(uint8_t                                bcc,
 
     m_timer_value_on_radio_end_event = delta_time;
 
+    #if ENABLE_ANT_DIV
+    // Select antenna
+    nrf_802154_trx_antenna_update();
```
ankuns: I think that such call is missing in `nrf_802154_trx_enable`. When exiting from `TRX_STATE_DISABLED` to `TRX_STATE_IDLE` the state of radio, and related transceiver circuits (including fem, ant div) is undefined, as you never know the state in which an other driver working outside out timeslot left those resources. 

Please also dispel doubts if `nrf_802154_trx_antenna_update` should or should not be called also in transmit frame, continuous carrier, etc... Currently there is a call to `nrf_802154_trx_antenna_update` in nrf_802154_trx_energy_detection, and here.

src/mac_features/nrf_802154_delayed_trx.c
```diff
@@ -377,7 +378,14 @@ static void rx_timeslot_started_callback(rsch_dly_ts_id_t dly_ts_id)
...
+            update_result = nrf_802154_request_channel_update() &&
+                            nrf_802154_request_antenna_update();
```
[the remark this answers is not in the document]
ankuns: Secondly calling nrf_802154_request_antenna_update will at the far end call `nrf_802154_trx_antenna_update` which is called again when trx enters receive mode (see `nrf_802154_trx_receive_frame`). Is there any reason for doing this operation twice?

src/nrf_802154_types.h
JakubPegza: Antenna is not updated by TRX if in disabled mode. This would allow possibility of giving the control of the antenna away to any other module (through GPIO). 
On switching to disabled mode, however, antenna is switched to default antenna (that is, antenna that would be used if no ant_div module was present), which is antenna 1. This makes sure the antenna is always in the same state when switching to disabled mode.

src/nrf_802154_trx.c
JakubPegza: Antenna diversity is independent from fem, although I agree that control of the antenna should be ceded for the same reasons.

src/nrf_802154_trx.c
JakubPegza: Currently, antenna diversity is supposed to be only used in reception, that's why no calls to update are made before transmission, continuous carrier etc. 
However,  tx_antenna update and rx_antenna update probably should be separated, with tx_antenna_update being stub for now (working as disabled if ant_div is disabled and working as ant1 in any other case). This could prevent using random (or, more precisely, the same as for previous reception) antenna for transmission when the antenna_diversity is enabled.

I'm not sure about call to update in TRX_STATE_DISABLED->TRX_STATE_IDLE transmission in that case. Antenna still would be updated before use (before receive and in future also transmit), but with ant_div for tx and rx being different, it is unclear which should be used in TRX_STATE_IDLE.

src/mac_features/nrf_802154_delayed_trx.c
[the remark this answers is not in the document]
JakubPegza: This was intentional, as rx request would never be called if channel update failed. However, I think you're right that it may be unnecessary, as it would be called anyway after entering receive mode.

src/nrf_802154_types.h
ankuns: So let such explanation be in dox.

src/nrf_802154_types.h
```diff
@@ -205,6 +205,34 @@ typedef uint8_t nrf_802154_ifs_mode_t;
...
+#define NRF_802154_ANT_DIV_MODE_ANTENNA_1 0x01 // !< First antenna is used
+#define NRF_802154_ANT_DIV_MODE_ANTENNA_2 0x02 // !< Second antenna is used
+#define NRF_802154_ANT_DIV_MODE_MANUAL    0x03 // !< Antenna is selected manually
```
jciupis: As discussed F2F, there's no actual need for `ANTENNA_1` and `ANTENNA_2` modes. They can be replaced with `MANUAL` mode and PIB configuration.

src/nrf_802154.c
```diff
@@ -301,6 +301,52 @@ void nrf_802154_fem_control_cfg_get(nrf_802154_fem_control_cfg_t * p_cfg)
...
+#if ENABLE_ANT_DIV
+
+bool nrf_802154_ant_div_mode_set(nrf_802154_ant_div_mode_t mode)
+{
+    bool result = nrf_802154_pib_ant_div_mode_set(mode);
+
+    if (result)
+    {
+        nrf_802154_request_antenna_update();
+    }
```
jciupis: Newline missing

src/mac_features/nrf_802154_delayed_trx.c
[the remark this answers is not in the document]
jciupis: +1 to the `nrf_802154_request_antenna_update` being obsolete here

src/nrf_802154_ant_div.h
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * @brief Retrieves the antenna diversity configuration.
+ *
+ * @return Current antenna diversity module configuration.
+ */
+nrf_802154_ant_div_config_t nrf_802154_ant_div_config_get(void);
+
+/**
+ * @brief Select an antenna to use.
```
jciupis: Please use 3rd person verbs in the documentation as agreed upon with the techwriter: `Selects`

src/nrf_802154_ant_div.c
jciupis: +1 to move it to `init()` function. This structure will most probably grow and no one will remember to care about it then, it's better to do it straight away IMO

JakubPegza: After F2F discussion, initialization will be left outside of `_init()` function. Reason being, `_init()` function is to be called after reading and writing config, which requires it to be already initialized and not overwritten by init.

JakubPegza:
Post-review fixes, minor refactor and rebase onto Nordic repository feature/antenna_diversity

src/nrf_802154_trx.c
```diff
@@ -48,6 +48,9 @@
 #include "nrf_802154_procedures_duration.h"
 #include "nrf_802154_critical_section.h"
 #include "fem/nrf_fem_protocol_api.h"
+#if ENABLE_ANT_DIVERSITY
```
jciupis: This check is probably unnecessary

src/nrf_802154_trx.c
```diff
@@ -695,6 +698,92 @@ void nrf_802154_trx_disable(void)
...
+    nrf_802154_ant_div_mode_t mode   = nrf_802154_pib_ant_div_mode_get();
+
+    switch (mode)
+    {
+        case NRF_802154_ANT_DIV_MODE_DISABLED:
+            /* Intentionally empty. */
+            break;
+
+        case NRF_802154_ANT_DIV_MODE_MANUAL:
+            result = nrf_802154_ant_div_antenna_set(NRF_802154_ANT_DIV_ANTENNA_1);
```
jciupis: `NRF_802154_ANT_DIV_ANTENNA_1` --> `nrf_802154_pib_ant_div_antenna_get()` ?

src/nrf_802154_trx.c
```diff
@@ -695,6 +698,92 @@ void nrf_802154_trx_disable(void)
...
+        case TRX_STATE_STANDALONE_CCA:
+        case TRX_STATE_RXACK:
+        case TRX_STATE_TXFRAME:
+        case TRX_STATE_CONTINUOUS_CARRIER:
+        case TRX_STATE_MODULATED_CARRIER:
+            tx_antenna_update();
+            break;
+
+        case TRX_STATE_DISABLED:
+            assert(false);
```
jciupis: One assert for that is enough ;)

src/nrf_802154_pib.h
```diff
@@ -319,6 +319,50 @@ uint16_t nrf_802154_pib_ifs_min_lifs_period_get(void);
...
+
+/**
+ * @brief Gets the antenna diversity mode.
+ *
+ * @return Current antenna diversity mode.
+ */
+nrf_802154_ant_div_mode_t nrf_802154_pib_ant_div_mode_get(void);
+
+/**
+ * @brief Select antenna to be used.
```
jciupis: Use third person verbs in the `@brief` sections

src/nrf_802154_pib.c
```diff
@@ -46,7 +46,11 @@
 #include "nrf_802154_utils.h"
 #include "fal/nrf_802154_fal.h"
 
-#define CSMACA_BE_MAXIMUM 8 ///< The maximum allowed CSMA-CA backoff exponent (BE) that results from the implementation
+#if ENABLE_ANT_DIVERSITY
```
jciupis: I'm not really convinced by these conditional includes. IMO it'd be better to move these preprocessor directives to the header / source file as it's done with other conditionally compiled features

src/nrf_802154_ant_diversity.h
```diff
@@ -0,0 +1,96 @@
...
+ *
+ * @param[in] antenna  Antenna to be used.
+ *
+ * @retval true  Antenna switched successfully.
+ * @retval false Invalid antenna passed to the function.
+ */
+bool nrf_802154_ant_div_antenna_set(nrf_802154_ant_div_antenna_t antenna);
+
+/**
+ * @brief Get currently used antenna.
```
jciupis: 3rd person verb, please :)

src/nrf_802154.h
```diff
@@ -174,6 +178,77 @@ void nrf_802154_fem_control_cfg_get(nrf_802154_fem_control_cfg_t * p_cfg);
 
 #endif // ENABLE_FEM
 
+#if ENABLE_ANT_DIVERSITY
+/**
+ * @brief Set the antenna diversity mode
+ *
+ * Sets the mode of the antenna diversity module. Should only be called
```
jciupis: The first sentence repeates the brief note and is therefore redundant. What do you think about:
```
* @brief Sets the antenna diversity mode.
*
* @note This function should only be called when TRX is disabled.
```

src/nrf_802154.c
```diff
@@ -301,6 +301,58 @@ void nrf_802154_fem_control_cfg_get(nrf_802154_fem_control_cfg_t * p_cfg)
...
+nrf_802154_ant_div_mode_t nrf_802154_ant_div_mode_get(void)
+{
+    return nrf_802154_pib_ant_div_mode_get();
+}
+
+bool nrf_802154_antenna_set(nrf_802154_ant_div_antenna_t antenna)
+{
+    bool result = nrf_802154_pib_ant_div_antenna_set(antenna);
+
+    if ((result) && (NRF_802154_ANT_DIV_MODE_MANUAL == nrf_802154_pib_ant_div_mode_get()))
```
jciupis: With this approach, if we're not in the manual mode, we can end up having incorrect information in PIB. Please perform this check earlier

src/nrf_802154_trx.c
JakubPegza: It is supposed to be a stub for now, as antenna diversity for tx is not yet supported by this module, therefore, the less impact it has on antenna during transmission (i.e. always selects default antenna before transmitting), the better.

jciupis: Probably worth a comment then.

src/nrf_802154.c
JakubPegza: As stated in `nrf_802154_antenna_get` function documentation, pib is not supposed to reflect the state of the antenna outside of manual mode. This allows to set the antenna while manual mode is disabled, and the antenna used will be updated after entering manual mode. Possible use case could be different modules controlling the antenna in different timeslots, where antenna diversity module would be disabled outside of timeslot and enabled in specified (in this case - manual) state when in timeslot. Antenna could be then set to the specified value, and enabling the antenna diversity module would update the real antenna state. When in disabled mode however (antenna diversity module cedes the control of the antenna), value set in pib does not reflect the real state of the antenna.
Do you think this is design error? Should it be stated more clearly in the documentation?

jciupis: I agree with the explanation completely, thanks for the clarification.

The documentation in PIB is quite clear about `nrf_802154_pib_ant_div_antenna_get` not being accurate outside of manual mode. The public API (`nrf_802154_antenna_get`) documentation could state it clearly as well.

PR #330, opened 2019-12-09
Cast explicitly to PPI & TIMER HAL types

jciupis:


jciupis:
Regression test

hubertmis:
Continuous integration missed single frame during the run. Force merging to speed up release process

PR #334, opened 2019-12-18
Fix TX termination procedure

jciupis:


src/nrf_802154_trx.c
```diff
@@ -1323,11 +1324,24 @@ static inline void wait_until_radio_is_disabled(void)
...
-        /* Intentionally empty */
+        if (nrf_radio_state_get() == NRF_RADIO_STATE_DISABLED)
+        {
+            radio_is_disabled = true;
+            break;
+        }
     }
 
+    assert(radio_is_disabled);
+
```
ankuns: here possibly `(void)radio_is_disabled;` When NDEBUG is build compiler might argue about value set not used. Please just check.

jciupis: Nice catch! It'd surely generate a compilation error, I'm fixing it immediately.

PR #336, opened 2019-12-20
Clear shared PPI channel when the driver stops using it.

hubertmis:
There was a problem with FEM initialization if the driver left some data
in disabled PPI channel. Now PPI channel is cleared when the driver ends
using it.


src/nrf_802154_trx.c
```diff
@@ -1371,6 +1373,9 @@ static void rxframe_finish_disable_ppis(void)
     nrf_ppi_channel_disable(PPI_CRCOK_DIS_PPI);
     nrf_ppi_channel_disable(PPI_ADDRESS_COUNTER_COUNT);
     nrf_ppi_channel_disable(PPI_CRCERROR_COUNTER_CLEAR);
+
+    nrf_ppi_channel_endpoint_setup(PPI_CRCERROR_CLEAR, 0, 0);
+    nrf_ppi_fork_endpoint_setup(PPI_CRCERROR_CLEAR, 0);
```
jciupis: It seems that depending on the driver's configuration, only `PPI_EGU_TIMER_START` and `PPI_EGU_RAMP_UP` use their forks and need to have them cleared. Please double-check - that would make clearing these forks obsolete.

hubertmis:
Test results: http://projecttools.nordicsemi.no/bamboo/browse/KNG-RV8MB10-3
Fail rate exactly like int/wifi branch. No regression spotted. Merging

PR #339, opened 2020-01-14
Align DTX timing constants to latest measurements

jciupis:


hubertmis:
@jciupis , have you tried it with all compilers we support?

jciupis:
@hubertmis With all the compilers that are supported by the int/wifi branch, yes.

PR #346, opened 2020-01-27
Automatic mode of antenna diversity module

JakubPegza:
Recreate PR with antenna diversity PIB configuration after major rework.
Implement automatic switcher mode in two variants.


JakubPegza:
Post-rebase artifacts left, need to rebase and reopen on a clean branch.

PR #347, opened 2020-01-27
Feature/ant div auto

JakubPegza:
Recreate previous antenna diversity PRs after major rework and clean rebase.
Implement automatic switcher in 2 variants, rework antenna diversity module structure, expand API with toggle timing configuration.


src/nrf_802154.c
```diff
@@ -269,6 +269,75 @@ void nrf_802154_fem_control_cfg_get(nrf_802154_fem_control_cfg_t * p_cfg)
...
+    nrf_802154_ant_diversity_config_set(config);
+
+}
+
+nrf_802154_ant_diversity_antenna_t nrf_802154_last_rx_antenna_get(void)
+{
+    return nrf_802154_ant_diversity_last_rx_antenna_get();
+}
+
+void nrf_802154_ant_diversity_toggle_time_set(uint8_t toggle_time)
```
czeslawmakarski: Why `toggle_time` `uint8_t` but the corresponding getter returns `uint32_t`?

src/nrf_802154_ant_diversity.c
```diff
@@ -0,0 +1,620 @@
...
+/**
+ * @file
+ *   This file implements the 802.15.4 antenna diversity module.
+ *
+ */
+#define NRF_802154_MODULE_ID NRF_802154_MODULE_ID_AD
+
+#include <assert.h>
+
+#include "nrf_802154_ant_diversity.h"
```
czeslawmakarski: AFAIK this include should be the first one.

src/nrf_802154_pib.c
```diff
@@ -492,3 +517,71 @@ void nrf_802154_pib_ifs_min_lifs_period_set(uint16_t period)
...
 #endif // NRF_802154_IFS_ENABLED
+
+#if ENABLE_ANT_DIVERSITY
+bool nrf_802154_pib_ant_diversity_mode_set(nrf_802154_ant_diversity_mode_t mode)
+{
+    bool result = true;
+
+    switch (mode)
+    {
+        case NRF_802154_ANT_DIVERSITY_MODE_DISABLED:
```
jciupis: I think that upon entering the disabled mode, antenna should be changed to `NONE`

src/nrf_802154.c
```diff
@@ -269,6 +269,75 @@ void nrf_802154_fem_control_cfg_get(nrf_802154_fem_control_cfg_t * p_cfg)
 
 #endif // ENABLE_FEM
 
+#if ENABLE_ANT_DIVERSITY
+bool nrf_802154_ant_diversity_mode_set(nrf_802154_ant_diversity_mode_t mode)
```
jciupis: It's misleading that this function has the same prefix (`nrf_802154_ant_diversity`) as the ant_diversity module API as if it implemented part of that interface. Please rename it to avoid misunderstandings.

src/nrf_802154.c
```diff
@@ -269,6 +269,75 @@ void nrf_802154_fem_control_cfg_get(nrf_802154_fem_control_cfg_t * p_cfg)
...
+{
+    return nrf_802154_pib_ant_diversity_toggle_time_get();
+}
+
+nrf_802154_ant_diversity_config_t nrf_802154_antenna_config_get(void)
+{
+    return nrf_802154_ant_diversity_config_get();
+}
+
+void nrf_802154_antenna_diversity_init(void)
```
jciupis: Please unify the functions' names and make it clear that they're a part of public API that is completely separate from the ant_diversity module API. Currently we have `ant_diversity`,`antenna`, `antenna_diversity`. Difficult to follow

src/nrf_802154_ant_diversity.h
```diff
@@ -0,0 +1,173 @@
...
+#include <stdint.h>
+
+#include "nrf_802154_types.h"
+
+#ifndef NRF_802154_ANT_DIVERSITY_ANT_SEL_PIN_DEFAULT
+#define NRF_802154_ANT_DIVERSITY_ANT_SEL_PIN_DEFAULT 23
+#endif
+
+#ifndef RSSI_SETTLE_TIME_US
+#define RSSI_SETTLE_TIME_US 16
```
jciupis: It'd be nice to have just a tiny bit of documentation what this macro represents

src/nrf_802154_ant_diversity.h
```diff
@@ -0,0 +1,173 @@
...
+void nrf_802154_ant_diversity_init(void);
+
+/**
+ * @brief Selects an antenna to use.
+ *
+ * @note This function has no effect while antenna diversity module is currently
+ * toggling antenna in PPI variant - that is:
+ *  - Antenna diversity is in PPI variant and auto mode is enabled
+ *  - rx is enabled
+ *  - no PPDU is currently being received
```
jciupis: `PSDU`?

src/nrf_802154_ant_diversity.h
```diff
@@ -0,0 +1,173 @@
...
+ *
+ * @retval true  Antenna switched successfully.
+ * @retval false Invalid antenna passed to the function.
+ */
+bool nrf_802154_ant_diversity_antenna_set(nrf_802154_ant_diversity_antenna_t antenna);
+
+/**
+ * @brief Gets currently used antenna.
+ *
+ * @return Currently used antenna.
```
jciupis: Does it actually return the currently used antenna or does it have the same caveats as e.g. the PIB implementation?

src/nrf_802154_ant_diversity.c
```diff
@@ -0,0 +1,620 @@
...
+#include "nrf_802154_debug.h"
+#include "nrf_gpio.h"
```
jciupis: Please follow an alphabetical order of includes

src/nrf_802154_ant_diversity.c
```diff
@@ -0,0 +1,620 @@
...
+
+        case AD_STATE_SLEEP:
+        case AD_STATE_TOGGLE:
+        case AD_STATE_SETTLE_1:
+        case AD_STATE_SETTLE_2:
+            // Intentionally empty
+            break;
+
+        default:
+            assert(false);
```
jciupis: missing break here

src/nrf_802154.h
```diff
@@ -174,6 +178,107 @@ void nrf_802154_fem_control_cfg_get(nrf_802154_fem_control_cfg_t * p_cfg);
...
+nrf_802154_ant_diversity_config_t nrf_802154_antenna_config_get(void);
+
+/**
+ * @brief Sets the antenna toggle time in automatic mode.
+ *
+ * See @ref nrf_802154_ant_diversity_mode_set.
+ *
+ * @param[in] toggle_time  Time between switching antennas in us.
+ */
+void nrf_802154_ant_diversity_toggle_time_set(uint8_t toggle_time);
```
hubertmis: It looks that this function is implementation-specific. If we replaced implemented algorithm with other that does not toggle in constant intervals, this function could not be reused. Would it be possible to merge it with `nrf_802154_antenna_config_set()`, or change to `nrf_802154_ant_diversity_config_set(union or void*)` to reuse the same API for different algorithms if we decide to create new?

src/nrf_802154_peripherals.h
```diff
@@ -106,6 +106,70 @@ extern "C" {
...
+#define NRF_802154_ANT_DIVERSITY_TIMER_INSTANCE_NO 3
+#endif
+
+/**
+ * @def NRF_802154_COUNTER_TIMER_INSTANCE
+ *
+ * The timer instance used by the driver for detecting when PSDU is being received.
+ *
+ * @note This configuration is used only when the NRF_RADIO_EVENT_BCMATCH event handling is disabled
+ *       (see @ref NRF_802154_DISABLE_BCC_MATCHING).
```
ankuns: Please write clear, This configuration is used only when NRF_802154_DISABLE_BCC_MATCHING == (use required value).

src/nrf_802154_trx.c
```diff
@@ -2399,7 +2547,7 @@ void nrf_802154_radio_irq_handler(void)
     // That's why we check here EGU's EGU_HELPER1_INTMASK.
     // The RADIO does not have interrupt from HELPER1 event.
     if (nrf_egu_int_enable_check(NRF_802154_SWI_EGU_INSTANCE, EGU_HELPER1_INTMASK) &&
-        nrf_radio_event_check(NRF_RADIO_EVENT_HELPER1))
+        nrf_radio_event_check(NRF_RADIO_EVENT_HELPER1) && (m_trx_state == TRX_STATE_RXFRAME))
```
ankuns: Just please make sure this is not a leftover. I think that this additional condition should not be checked, because it may lead to not clearing event and periodical entering into irq handler.

src/nrf_802154_pib.c
JakubPegza: I don't think it should. The idea of disabled mode is to cede control of the antenna to another module (possibly for limited time), or to stop using antenna diversity altogether. The way it is implemented right now, going back from disabled to manual mode, antenna diversity module remembers which antenna was selected previously and selects it next time RX is enabled. If `NONE` would be left, the antenna would have to be reselected each time (so, the higher layer would have to store it somewhere), or the module would have to either 
 * resort to `DEFAULT` antenna, which you mentioned in other discussion,
 * or act as if still in `DISABLED` mode. 

Both of these options seem unintuitive to me, but I'm open to further arguments.

I could argue though, that adapting the idea above could call for splitting `MODE` selection (manual/auto) and disabling/enabling the module, so that even the last selected mode is remembered when coming back from `DISABLED` state. This way, initialization would be done by calling 
`mode_set()`
`ant_div_enable()`
but coming back from `DISABLED` state, only `ant_div_enable` would be required, instead of higher layer having to remember the mode used as well.
I think this would, however, cause API to grow unnecessarily, while not providing significant improvement in usability, so I would not change that right now.

src/nrf_802154_ant_diversity.h
JakubPegza: `PPDU` is correct here, as toggling the antenna ends as soon as preamble is detected - around 120 us before PSDU starts.

src/nrf_802154_core.c
```diff
@@ -1402,40 +1407,77 @@ static void on_rx_prestarted_timeout(void * p_context)
...
+        rx_timeout_should_be_started = true;
+    }
+
+    /*
+     * This handler might not be followed by nrf_802154_trx_receive_frame_started. The timer
+     * below is used for timing out if the framestart doesn't come.
+     * There are two reasons for that: reverting boosted preconditions and notifying antenna diversity
+     * module.
+     */
+    if (rx_timeout_should_be_started)
```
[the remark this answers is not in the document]
jciupis: I forgot about the counters also using this interrupt. Consider this comment solved

src/nrf_802154_pib.c
jciupis: Fair enough. I'm OK to leave it as it is.

src/nrf_802154_ant_diversity.h
JakubPegza: No caveats here, this function reads the state of the actual pin.

src/nrf_802154.h
JakubPegza: It seems like a reasonable idea, but I can see one caveat - `ant_diversity_config` is the config stored in antenna diversity module, not in pib, as it is responsible for the `ant_sel` pin. 
Unifying those two functions would require the toggle time to be moved from pib to antenna diversity, but, as toggle time is implementation-specific, I see no problem in that.

src/nrf_802154_ant_diversity.h
hubertmis: I would move it to nrf_802154_const.h - it's not only antenna diversity related

src/nrf_802154_ant_diversity.h
```diff
@@ -0,0 +1,173 @@
...
+
+#ifndef NRF_802154_ANT_DIV_H
+#define NRF_802154_ANT_DIV_H
+
+#include <stdint.h>
+
+#include "nrf_802154_types.h"
+
+#ifndef NRF_802154_ANT_DIVERSITY_ANT_SEL_PIN_DEFAULT
+#define NRF_802154_ANT_DIVERSITY_ANT_SEL_PIN_DEFAULT 23
```
hubertmis: It seems this value should be included in nrf_802154_peripherals.h in the `NRF_802154_GPIO_PINS_USED_MASK` macro

src/nrf_802154_ant_diversity.c
```diff
@@ -0,0 +1,620 @@
...
+#endif // ANT_DIVERSITY_PPI
+
+void nrf_802154_ant_diversity_init(void)
+{
+    ad_timer_init();
+#if ANT_DIVERSITY_PPI
+    ad_ppi_and_gpiote_init();
+#endif // ANT_DIVERSITY_PPI
+
+    nrf_gpio_cfg_output(m_ant_div_config.ant_sel_pin);
```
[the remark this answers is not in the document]
JakubPegza: GPIOTE task for that pin is not yet enabled here, it's only enabled after entering `TOGGLE` state. Only initial configuration is performed here. Besides, in ppi mode, pin control is passed back and forth between GPIOTE and GPIO modules:
 * for toggling - GPIOTE
 * for RSSI and final antenna selection - GPIO

This responsibility change is mentioned in `ad_timer_toggle_deconfigure` and 'nrf_802154_ant_diversity_antenna_get`, as it's not really relevant anywhere else.

src/nrf_802154_ant_diversity.h
```diff
@@ -0,0 +1,173 @@
...
+/**
+ * @brief Switches the antenna currently in use.
+ *
+ * @note This function has no effect while antenna diversity module is currently
+ * toggling antenna in PPI variant - that is:
+ *  - Antenna diversity is in PPI variant and auto mode is enabled
+ *  - rx is enabled
+ *  - no PPDU is currently being received
+ */
+void nrf_802154_ant_diversity_antenna_toggle();
```
hubertmis: It seems that this function is an internal function of antenna diversity module and should be removed from the header file

src/nrf_802154_ant_diversity.c
```diff
@@ -0,0 +1,620 @@
...
+#if ANT_DIV_PPI
+    // While toggling via PPI, GPIOTE is responsible for ant_sel pin
+    // Different register has to be used for reading the pin state.
+    if (AD_STATE_TOGGLE == m_ad_state)
+    {
+        return nrf_gpio_pin_read(m_ant_div_config.ant_sel_pin);
+    }
+#endif // ANT_DIV_PPI
+    return nrf_gpio_pin_out_read(m_ant_div_config.ant_sel_pin);
+}
```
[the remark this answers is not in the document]
JakubPegza: If pin is configured as output, it's state is read from different register (`GPIO->OUT` instead of `GPIO->IN`) that's why different nrfx functions are called here depending on the pin configuration (`nrf_gpio_pin_out_read` and `nrf_gpio_pin_read` respectively). Unless there is a function that handles both cases that I'm not aware of?

src/nrf_802154_ant_diversity.c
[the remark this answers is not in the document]
ankuns: ok

src/nrf_802154_ant_diversity.h
```diff
@@ -0,0 +1,173 @@
...
+nrf_802154_ant_diversity_antenna_t nrf_802154_ant_diversity_antenna_get(void);
+
+/**
+ * @brief Gets which antenna was selected as best for the last reception.
+ *
+ * @note In three cases @ref NRF_802154_ANT_DIVERSITY_ANTENNA_NONE may be returned:
+ *  - No frame was received yet.
+ *  - Last frame was not received with antenna diversity auto mode enabled.
+ *  - RSSI measurements didn't have enough time to finish during last frame reception
+ *    and antenna was selected at random.
```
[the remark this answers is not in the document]
JakubPegza: I don't believe we have the technology yet for implementing true randomness :wink:
It's unpredictable, because in this case state of the antenna depends on the state in which the preamble was detected (which is constantly toggling between ant1 and ant2). After that, there is another consideration of whether zero or one RSSI measurements completed since prestarted, so antenna could be toggled one more time after that. Formula for this could be represented as
`((time_in_us_since_RXEN / toggle_time_us) % 2) ^ initial_state ^ (rssi_measurements_completed_since_prestarted % 2)`
This seems random enough for me, especially since the toggle time has limited precision as well.

src/nrf_802154_peripherals.h
JakubPegza: This was a leftover from copypasting the documentation-define format, fixed now (in this timer, original documentation is out of scope for this pr).

src/nrf_802154_ant_diversity.h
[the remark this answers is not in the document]
ankuns: I think that the word "unspecified" perfectly does the job in this case. Consider
```
 *  - RSSI measurements didn't have enough time to finish during last frame reception
 *    and it is unspecified which antenna was selected.
```
It simply means that antenna was selected **somehow**, the behavior may or may not be repeatable, but there is nothing to do with randomness here. It also means that the user should not rely on this.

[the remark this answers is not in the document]
JakubPegza: Fair enough, I'll correct that

src/nrf_802154_ant_diversity.h
JakubPegza: This pin is not always used, as the ant_sel pin could (and probably should) be reconfigured run-time, and only defaults to this if the configuration is not changed. In that case I'm not sure if it should really be included in `NRF_802154_GPIO_PINS_USED_MASK`. The pin could be configured entirely compile-time, but this leaves possibility of future precompiled releases without reconfiguration option. 
Nevertheless, this macro could be moved to `nrf_802154_peripherals.h` or `nrf_802154_config.h` (either way, none of those modules currently define any pins used), I'm not sure what is the best place for this, any thoughts?

hubertmis: I didn't think about it so deeply. Perhaps we can leave it as it is right now and rethink pins config during design of the new library APIs.

src/nrf_802154_peripherals.h
```diff
@@ -106,6 +106,70 @@ extern "C" {
...
+#define NRF_802154_ANT_DIVERSITY_GPIOTE_CHANNELS_USED_MASK (1 << \
+                                                            NRF_802154_ANT_DIVERSITY_GPIOTE_CHANNEL)
+
+#endif // ANT_DIVERSITY_PPI
+
+#else  // ENABLE_ANT_DIVERSITY
+
+#define NRF_802154_ANT_DIVERSITY_TIMER_MASK 0
+
+#endif // ENABLE_ANT_DIVERSITY
```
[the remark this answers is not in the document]
JakubPegza: Macros for PPI and GPIOTE channels used mask should be set to 0 in two cases:
 * antenna diversity is compiled in SW variant
 * antenna diversity is not compiled into driver

In current form, those macros need to be defined only once. In case they were moved into `#else` block of `#if ENABLE_ANT_DIVERSITY`, they would need to be duplicated into `#else` block of `#if ANT_DIVERSITY_PPI` as well. 
All things considered, current implementation is a bit cleaner, don't you agree?

src/nrf_802154_ant_diversity.c
```diff
@@ -0,0 +1,639 @@
...
+    nrf_timer_int_enable(ANT_DIV_TIMER, NRF_TIMER_INT_COMPARE0_MASK);
+
+    NVIC_SetPriority(NRF_802154_ANT_DIVERSITY_TIMER_IRQN, 1);
+    NVIC_ClearPendingIRQ(NRF_802154_ANT_DIVERSITY_TIMER_IRQN);
+    NVIC_EnableIRQ(NRF_802154_ANT_DIVERSITY_TIMER_IRQN);
+#elif ANT_DIVERSITY_PPI
+    nrf_gpiote_task_enable(NRF_802154_ANT_DIVERSITY_GPIOTE_CHANNEL);
+    nrf_ppi_channel_enable(ANT_DIV_PPI);
+#else
+    assert(false);
```
ankuns: Here `#error Unsupported configuration` would be better than `assert`. We would get knowledge about pointless macro configuration at runtime which is far too late. We could be notified at compile time, couldn't we?

src/nrf_802154_ant_diversity.c
```diff
@@ -0,0 +1,639 @@
...
+    __DSB();
+    __ISB();
+#elif ANT_DIVERSITY_PPI
+    // Set GPIO output pin value to the same as currently set by GPIOTE module
+    // This prevents pin switching after control of the pin is ceded by GPIOTE
+    nrf_gpio_pin_write(m_ant_div_config.ant_sel_pin, nrf_802154_ant_diversity_antenna_get());
+    nrf_gpiote_task_disable(NRF_802154_ANT_DIVERSITY_GPIOTE_CHANNEL);
+    nrf_ppi_channel_disable(ANT_DIV_PPI);
+#else
+    assert(false);
```
ankuns: As for line 226 nearby

src/nrf_802154_types.h
```diff
@@ -205,6 +205,53 @@ typedef uint8_t nrf_802154_ifs_mode_t;
...
+#endif
+
+/**
+ * @brief Configuration of the antenna diversity module.
+ *
+ */
+typedef struct
+{
+    uint8_t ant_sel_pin; // !< Pin used for antenna selection.
+    // !< Should not be changed after calling @ref nrf_802154_ant_diversity_init.
```
jciupis: I have noticed that you placed such comments in various chunks of code:
```c
type_1_t var_1_name; // First line of the comment
// Second line of the comment
type_2_t var_2_name; // Third line of the comment
```

Depending on types and names of the variables, it can be quite difficult to tell which variable the middle comment line refers to. I'd suggest to either change it like this:
```c
type_1_t var_1_name; // First line of the comment
                     // Second line of the comment
type_2_t var_2_name; // Third line of the comment
```

or like this:
```c
// First line of the comment. Second line of the comment
type_1_t var_1_name;

// Third line of the comment
type_2_t var_2_name;
```

or whatever you find clean and unambiguous.

src/nrf_802154_peripherals.h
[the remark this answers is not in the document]
jciupis: I do.

src/nrf_802154_types.h
JakubPegza: My original idea was the first style you proposed, however, for some reason, this did not satisfy uncrustify. I'll try to format it in some other way

PR #260, opened 2019-07-31
Reworked FEM API

czeslawmakarski:
Introduce a new FEM API.

PR #354, opened 2020-05-08
port use of nrf_error.h to nrfx_errors.h as provided by upstream nrfx repo

dpiegdon:
In https://github.com/NordicSemiconductor/nRF-IEEE-802.15.4-radio-driver/pull/260 use of nrf_error.h and "NRF_SUCCESS" / "NRF_ERROR_*" macros was introduced.
but the official nrfx repository does use slightly different names.
this patch adapts the naming to work with the official nrfx repo.

please note that the changes were not tested extensively. they compile for me (tm), but you should run the internal unit tests. I was unable to figure out how to run them. it would be great to have a test-script to run these automatically.

### DIFF
--- a/src/fem/nrf_fem_control_config.h
+++ b/src/fem/nrf_fem_control_config.h
@@ -38,7 +38,7 @@
 #include <stdint.h>
 #include <stdbool.h>
 
-#include "nrf_error.h"
+#include "nrfx_errors.h"
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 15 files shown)


hubertmis:
I agree that using `NRF_SUCCESS` that is not defined in this repository as a return code was not a good idea. However, changing it to `NRFX_SUCCESS` is an API-breaker that should not be merged unless we change the major version of the driver.

Currently we are working internally to solve this problem. Stay tuned, an update will be added to this repository soon.

dpiegdon:
Thanks for the heads up, I agree.

PR #355, opened 2020-05-08
fix warning about several unused parameters

dpiegdon:
if a recent gcc with -Wunused-parameter is used, it complains about these.


dpiegdon:
unsure if I got *all* the unused parameters, as I only use a fraction of the source files.
maybe you can extend unit-tests to use a newer gcc and check for the appropriate warnings?

hubertmis:
We are working on the compiler warnings as well. We are in the middle of the process of CI build update. Because of that you see failures on each PR.
Expected update on this repository soon, then we'll discuss how to approach this PR.

PR #356, opened 2020-05-08
fix warning about invalid use of variadic macro without parameter

dpiegdon:
recent GCC complains that:
ISO C99 requires at least one argument for the "..." in a variadic macro

so use a separate macro for functions without parameters.


dpiegdon:
unsure if I got *all* the variadic template warinngs, as I only use a fraction of the source files.
maybe you can extend unit-tests to use a newer gcc and check for the appropriate warnings?

